PR #1, opened 2020-09-09
First round of alert stabilization

freddyrios:
- fixed: 123,2 was being interpreted as 1232 (thousands separator are not allowed anymore)
- fixed: alert message contained all values of all calls to CheckValue (now it only contains the last one)
- fixed: NaN and int were not allowed to be case insensitive as intended
- alert now reports wrong format if there is no comparison
- alert now reports wrong format if there is no value
- added initial set of unit tests for IOconfAlert (note it does not cover code in ServerUploader)


copenhagenatomics:
Looks good. But I would prefer this format in the IO.conf file  :
Alert;SomeValue < 202
No semicolons in the compare string but space as seperator. Ideally it would support both formats.

PR #11, opened 2020-09-10
Fixed #10: ServerUploader has issues handling Math

### LINKED ISSUE #10: ServerUploader has issues handling Math statements
```
            var dic = GetVectorDictionary(vector); //TODO: vector comes without the math values, yet this method will try to read those values as Math is in the descriptors
            IOconfFile.GetMath().ToList().ForEach(x => vector.Add(x.Calculate(dic)));
```

freddyrios:
- Introduced MathVectorExpansion and added unit tests for the math expansion.
- Fixed: vector expansion throws an exception trying to find math values before they are added
- The vector expansion now enforces the original vector length (instead of allowing extra math statements)

### DIFF
--- a/CA_DataUploaderLib/IOconf/IOconfMath.cs
+++ b/CA_DataUploaderLib/IOconf/IOconfMath.cs
@@ -24,8 +24,8 @@ namespace CA_DataUploaderLib.IOconf
             }
         }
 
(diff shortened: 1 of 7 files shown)


copenhagenatomics (approved):
Thanks for fixing this issue and adding unit tests.

Issue #6, opened 2020-09-09
alerts may fire too many times when recurrently crossing the condition boundary 


freddyrios:
If a value keeps hovering around the boundary condition of an alert, the same alert can trigger up to 150 times x minute (each 400 ms, more depending on the frequency of uploading measurements).

In the case of the is integer condition, it can trigger twice that amount when seeing repeated values. For this case, here are some cases that can be added to AlertDoesNotTriggersWhenOldValueMatched:
        [DataRow("Alert;MyName;int", 124d, 123d)]
        [DataRow("Alert;MyName;int", 123d, 122d)]
        [DataRow("Alert;MyName;int", 99999d, 100000d)]
        [DataRow("Alert;MyName;int", 9999999d, 10000000d)]
        [DataRow("Alert;MyName;int", 0d, 1d)]
        [DataRow("Alert;MyName;int", -2d, -1d)]
        [DataRow("Alert;MyName;int", -9999999d, -10000000d)]

freddyrios:
offline input: consider adding an extra argument to alerts that states the minimum minutes that need to pass before triggering again.

PR #13, opened 2020-09-10
Fixed #9: nan value causes alert to be skipped

### LINKED ISSUE #7: alerts are inconsistently triggering on the first measure
When (re)starting the program, some alerts trigger on the first measure and others don't.

This is caused because 0 is being used as the last recorded value on the first measure checked for each alert. This causes the > alerts to consider the boundary has been crossed, while the < to consider the opposite.

Instead, all alerts should trigger in the first measure if they match the condition.

For example, add these to IOconfAlertTests.AlertTriggers:
        [DataRow("Alert;MyName;<=;123", 123d)]
        [DataRow("Alert;MyName;<=;123", 122d)]
        [DataRow("Alert;MyName;<;123", 122d)]

### LINKED ISSUE #9: NaN causes the alert to be skipped
Receiving a NaN (not a number) value in the last value before the alert triggers can cause the alert to be skipped.

For example, add these to AlertTriggersWhenOldValueDidNotMatch
        [DataRow("Alert;MyName;>;123", double.NaN, 123.00012d)]
        [DataRow("Alert;MyName;>=;123", double.NaN, 123d)]
        [DataRow("Alert;MyName;<=;123", double.NaN, 122d)]
        [DataRow("Alert;MyName;!=;NaN", double.NaN, 122d)]

freddyrios:
Also Fixed #7

### DIFF
--- a/CA_DataUploaderLib/IOconf/IOconfAlert.cs
+++ b/CA_DataUploaderLib/IOconf/IOconfAlert.cs
@@ -63,24 +63,32 @@ namespace CA_DataUploaderLib.IOconf
         private double Value;
         private double LastValue;
         private string MessageTemplate;
+        private bool _isFirstCheck = true;
 
 
         public bool CheckValue(double newValue)
         {
-            Message = MessageTemplate + newValue.ToString(CultureInfo.InvariantCulture) + ")"; 
+            Message = MessageTemplate + newValue.ToString(CultureInfo.InvariantCulture) + ")";
             double lastValue = LastValue;
             LastValue = newValue;
-            switch(type)
+            var res = RawCheckValue(newValue) && (_isFirstCheck || !RawCheckValue(lastValue));
+            _isFirstCheck = false;
+            return res;
+
+            bool RawCheckValue(double val)
             {
-                case AlertCompare.EqualTo:  return newValue == Value && lastValue != Value;
-                case AlertCompare.NotEqualTo: return newValue != Value && lastValue == Value;
-                case AlertCompare.BiggerThan: return newValue > Value && lastValue <= Value;
-                case AlertCompare.SmallerThan: return newValue < Value && lastValue >= Value;
-                case AlertCompare.BiggerOrEqualTo: return newValue >= Value && lastValue < Value;
-                case AlertCompare.SmallerOrEqualTo: return newValue <= Value && lastValue > Value;
-                case AlertCompare.NaN: return Double.IsNaN(newValue) && !Double.IsNaN(lastValue);
-                case AlertCompare.IsInteger: return Math.Abs(newValue % 1) <= (Double.Epsilon * 100);
-                default: throw new Exception("IOconfAlert: this should never happen");
+                switch (type)
+                {
+                    case AlertCompare.EqualTo: return val == Value;
+                    case AlertCompare.NotEqualTo: return val != Value;
+                    case AlertCompare.BiggerThan: return val > Value;
+                    case AlertCompare.SmallerThan: return val < Value;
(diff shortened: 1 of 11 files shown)


freddyrios:
Note: to be rebased and merged to master after completing https://github.com/copenhagenatomics/CA_DataUploader/pull/11

copenhagenatomics (changes requested):
This new code is more difficult to read than before. Thus it is more likely to catch errors in the future. 
If "IsInteger" is the main reason behind the change. Then I suggest we delete it. I am not sure we will ever need it. Or at least move the RawCheckValue function outside the function as a private method.

freddyrios:
> This new code is more difficult to read than before. Thus it is more likely to catch errors in the future.
> If "IsInteger" is the main reason behind the change. Then I suggest we delete it. I am not sure we will ever need it. Or at least move the RawCheckValue function outside the function as a private method.

@copenhagenatomics the change was not due to the is integer issue mentioned at #6. In fact, it makes a poor job for that use case, because while it prevents a repeated number from triggering it does not trigger on different integers. That is, in the following sequence it only triggers twice, instead of 3 times: 
- 123.2
- 124
- 125
- 125.3
- 126

Instead, the change was done precisely to fix the referenced issues, that is:
- #9 
- #7 

Here are some ways of fixing NaN causing alerts to be skipped on the next value:
1. check if the last value was NaN and depending on the expression, the boundary value and the new measured value decide whether to skip
2. add is NaN checks to the different comparisons in the existing switch statement
3. use a negated comparison when comparing the old value in the existing switch statement i.e. return newValue > Value && !(newValue > Value)
4. same concept as previous point, but instead of spreading the negation in the switch statements, extract a method/local function and use it for both the match (newvalue) and no match checks (oldvalue).

Went with option 4, as 1 to 3 are more error prone and likely more complex in practice.

Regarding moving from a [local function](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/local-functions) to a private method, I take it then it is one of our code style choices to stick to private methods in general, right? Here is what the case that link makes for local functions:
>Local functions make the intent of your code clear. Anyone reading your code can see that the method is not callable except by the containing method. For team projects, they also make it impossible for another developer to mistakenly call the method directly from elsewhere in the class or struct.

CA_DataUploaderLib/IOconf/IOconfAlert.cs
```diff
@@ -63,24 +63,32 @@ public IOconfAlert(string row, int lineNum) : base(row, lineNum, "Alert")
...
         public bool CheckValue(double newValue)
         {
-            Message = MessageTemplate + newValue.ToString(CultureInfo.InvariantCulture) + ")"; 
+            Message = MessageTemplate + newValue.ToString(CultureInfo.InvariantCulture) + ")";
             double lastValue = LastValue;
             LastValue = newValue;
-            switch(type)
+            var res = RawCheckValue(newValue) && (_isFirstCheck || !RawCheckValue(lastValue));
+            _isFirstCheck = false;
+            return res;
```
freddyrios, line 76: @copenhagenatomics is it easier to read if we do this instead?

```
var res = _isFirstCheck ? RawCheckValue(newValue) : (RawCheckValue(newValue) && !RawCheckValue(oldValue))
_isFirstCheck = false;
return res;
```

or even:
```
var newMatched = RawCheckValue(newValue) ;
var res = _isFirstCheck ? newMatched : (newMatched && !RawCheckValue(oldValue))
_isFirstCheck = false;
return res;
```

freddyrios: closing this conversation as used a slightly different approach

PR #15, opened 2020-09-14
Closes #14 by adding a separate name parameter

### LINKED ISSUE #14: multiple alerts for the same measurement
we need to support multiple alerts for the same measurement.

Right now, the alert does not really have its own name, so its name needs to match the measurement it targets. 
Due to this, an error is displayed informing there are 2 with the same name if these 2 are added to the same configuration:
```
Alert;YellowProbe;>;28
Alert;YellowProbe;>;100
```

freddyrios:
Now the alerts take the shape: Alert;MyAlert;MySensor;>;28


copenhagenatomics (approved):
OK

PR #17, opened 2020-09-14
Fixed #16 detect misconfigured alerts

### LINKED ISSUE #16: detect missing sensor for alert in configuration
Currently, alerts that has a sensor missing in the configuration cause the program to abort and give a LINQ sequence error that makes hard to track that it is due to a misconfigured alert.

We need to add validation that checks if the sensor is missing for an alert in the configuration.

At the moment these types of rules are checked at IOconfFile.CheckRules. Note the following approach works for sensors that have a single descriptor that matches the sensor name in the config:
```
foreach (var alert in GetAlerts().Where(a => !Table.OfType<IOconfInput>().Any(x => x.Name == a.Sensor)))
  CALog.LogErrorAndConsoleLn(LogID.A, $"ERROR in {Directory.GetCurrentDirectory()}\\IO.conf:{Environment.NewLine} Alert: {alert.Name} points to missing sensor: {alert.Sensor}");
```

An example case where the above gives issues is oxygen sensors, as those are creating 3 descriptors that add different suffixes to the name.

freddyrios:
The validation is done at the ServerUploader level where we have all the actual descriptors that will be used.

If any misconfigured alerts are detected, we prevent program execution.

### DIFF
--- a/CA_DataUploaderLib/ServerUploader.cs
+++ b/CA_DataUploaderLib/ServerUploader.cs
@@ -7,11 +7,9 @@ using System.IO.Compression;
 using System.Linq;
 using System.Net.Http;
 using System.Net.Http.Headers;
-using System.Runtime.CompilerServices;
 using System.Security.Cryptography;
 using System.Text;
 using System.Threading;
-using System.Threading.Tasks;
 using System.Xml.Serialization;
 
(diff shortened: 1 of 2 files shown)


copenhagenatomics (approved):
Super.

PR #18, opened 2020-09-18
Feature/enhanced logging on mcu board serial errors

freddyrios:
- Enhanced logging on MCU board serial errors
- Fixed MCU Board read of the header (avoiding influence in logging of failed values later after board initialization)


copenhagenatomics (approved):
When you change console color in CALog.cs you probably have to use lock(someObject) { }  to avoid concurrency problems.

PR #20, opened 2020-09-21
Fixed #19 board reconnect not skipping the header

### LINKED ISSUE #19: board reconnection is not skipping the header
Box reconnection is not skipping the header and is not giving much time for the board to initialize.

freddyrios:
* It now gives more time during the reconnect for the board to respond.
* We now give up to 5 seconds to read the serial during board initialization on the initial setup (boxes that rely on responding to the Serial command tend to need more time for this i.e. because of the values lines being otherwise generated).


copenhagenatomics (approved):
SafeReopen() reads the SerialPrint lines from the MCU to try to avoid problems elsewhere in the code. I am not sure I like this approach. I would have liked this to be in the BaseSensorBox.cs. But if it solves the problem for now, we can always move this later.

PR #24, opened 2020-09-29
Logging additional info about sensor box responses

freddyrios:
Extra responses logging in BaseSensorBox:
- logs unhandled responses (non number lines)
- logs cycles where no data was received
- logs exception info when reading on the 10 retries

Logging extra info about reopening of the connection

### DIFF
--- a/CA_DataUploaderLib/BaseSensorBox.cs
+++ b/CA_DataUploaderLib/BaseSensorBox.cs
@@ -140,8 +140,11 @@ namespace CA_DataUploaderLib
                 {
                     foreach (var board in _boards)
                     {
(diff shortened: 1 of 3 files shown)


copenhagenatomics (approved):
This is somewhat debug code. When do you expect this to be removed again? One option is to keep it in a branch until it should die again.

freddyrios:
I might have over compensated, but we had errors that were not being logged
and that is problematic. We might still have.

One can enhance/sample/merge it in a way that avoids flooding the logs on
error/unexpected cases, but I would rather have the extra noise at first
and later on plan improvements as we see more of it in practice.

PR #26, opened 2020-10-28
failure conditions improvements

freddyrios:
- Fixed scenario where BaseSensorBox would keep trying to finish reading a partial line returned by a stalled board (for at least 200 seconds until the safelimit of MCUBoard kicks in)
- BaseSensorBox now logs unhandled board responses
- BaseSensorBox and HeatingController no longer try to reconnect multiple times to the same board
- Added BaseSensorBox.GetValues that has simple values that match debug configurations of GetVectorDescriptionItems
- SafeReopen no longer bubbles up the already logged exceptions. As this is already a reconnection due to errors action + it already tracks a retry limit, there is not much additional the calling code can do with it. We already hit a case where time outs exceptions of it were being mixed with something else.
- fixed unexpected switchboard log when starting
- fixed: alerts were using a login token that expired in 15 mins (used sliding expiration on alerts, so it only happened once there were 15 mins without alerts). Now signed alerts are used instead, like we do for vectors.
- Showing/logging extra info upload errors
- Fixed year of the alerts timestamps

### DIFF
--- a/CA_DataUploaderLib/BaseSensorBox.cs
+++ b/CA_DataUploaderLib/BaseSensorBox.cs
@@ -250,7 +250,7 @@ namespace CA_DataUploaderLib
                 {
                     CALog.LogErrorAndConsoleLn(LogID.A, $"{Title} stale value detected for port: {item.Input.Name}{Environment.NewLine}{msSinceLastRead} milliseconds since last read - closing serial port to restablish connection");
                     item.ReadSensor_LoopTime = 0;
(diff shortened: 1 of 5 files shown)


copenhagenatomics (approved):
Good. Did you test this new code with the "debug" setting in io.conf?

freddyrios:
> Good. Did you test this new code with the "debug" setting in io.conf?

@copenhagenatomics not really, been using normal mode, as I kind of want to know that it gives enough information. I can give it a shot next week.

PR #41, opened 2020-12-09
implemented #3 - added alert expression format

### LINKED ISSUE #3: Support alert format without semicolons
Looks good. But I would prefer this format in the IO.conf file  :
Alert;SomeValue < 202
No semicolons in the compare string but space as separator. Ideally it would support both formats.

_Originally posted by @copenhagenatomics in https://github.com/copenhagenatomics/CA_DataUploader/pull/1#issuecomment-689374568_

freddyrios:
Also implemented #6.


copenhagenatomics (approved):
OK to delete the integer stuff.

PR #49, opened 2021-01-06
Added sampling rates to IO config.

copenhagenatomics:
added a new setting in IO.conf
See my pull request. You need to 
1.)add this to the LoopControl main loop delay
2.)Add it to all the IO.conf files as you update next time
3.) Test it, merge it, find a good value, such that we upload vector with approx 2 Hz for most of our systems. And sample with LoopControl main loop with something like 6 - 9 Hz.

### DIFF
--- a/CA_DataUploaderLib/IOconf/IOconfFile.cs
+++ b/CA_DataUploaderLib/IOconf/IOconfFile.cs
@@ -38,6 +38,7 @@ namespace CA_DataUploaderLib.IOconf
             {
                 if (row.StartsWith("LoopName")) return new IOconfLoopName(row, lineNum);
                 if (row.StartsWith("Account")) return new IOconfAccount(row, lineNum);
+                if (row.StartsWith("SampleRates")) return new IOconfSamplingRates(row, lineNum);
                 if (row.StartsWith("Map"))    return new IOconfMap(row, lineNum);
                 if (row.StartsWith("Math")) return new IOconfMath(row, lineNum);
                 if (row.StartsWith("Alert")) return new IOconfAlert(row, lineNum);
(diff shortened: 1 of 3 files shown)


CA_DataUploaderLib/IOconf/IOconfFile.cs
```diff
@@ -106,6 +102,20 @@ public static ConnectionInfo GetConnectionInfo()
...
         }
 
+        public static string GetLoopName()
+        {
+            return ((IOconfLoopName)Table.Single(x => x.GetType() == typeof(IOconfLoopName))).Name;
+        }
+
+        public static int GetVectorUploadDelay()
+        {
+            return ((IOconfSamplingRates)Table.Single(x => x.GetType() == typeof(IOconfSamplingRates))).VectorUploadDelay;
```
freddyrios, line 112: suggest to have a default, like: 
```
public static int GetVectorUploadDelay() => 
   (Table.OfType<IOConfSamplingRates>().SingleOrDefault() ?? IOConfSamplingRates.Default).VectorUploadDelay
```
Where this is a static readonly field with the default values close to current behavior (1Hz for upload and 5 Hz for sampling).

CA_DataUploaderLib/ServerUploader.cs
```diff
@@ -258,7 +255,7 @@ private void LoopForever()
                     }
 
                     PrintBadPackagesMessage(false);
-                    Thread.Sleep(MillisecondsBetweenUpload);  // only send approx. one time per second. 
+                    Thread.Sleep(IOconfFile.GetVectorUploadDelay());
```
freddyrios, line 258: probably something for a separate PR, but this sleep should consider how time it already passed since the last upload i.e. to keep the desired frequency.

CA_DataUploaderLib/IOconf/IOconfSamplineRates.cs
```diff
@@ -0,0 +1,31 @@
...
+                VectorUploadDelay = (int)(1000 / VectorUploadFreq);
+            }
+            catch(Exception ex)
+            {
+                throw new Exception("IOconfMath: wrong format - expression: " + row, ex);
+            }
+        }
+
+        public int MainLoopDelay;
+        public int VectorUploadDelay;
```
freddyrios, line 29: We will need to update the latest graphs sub sampling logic that assumes we upload approx. once x second. To do this, we could have the server extract it from the vector description.

freddyrios (approved):

CA_DataUploaderLib/IOconf/IOconfFile.cs
freddyrios: @copenhagenatomics I have the impression the commit did not make it in?

PR #52, opened 2021-01-13
added loop control plugin support

freddyrios:
Commands added:
- load [auto]
- unload
- load myassembly.dll
- unload myassembly.dll

The commands allow to load/unload plugins. When no assembly is specified it loads all plugins from assembly files in the control program folder. If auto is specified, it automatically load/unload plugins as they are added/removed from the folder.

To implement a plugin a class must be created that derives from LoopControlPlugin and receives a CommandHandler constructor parameter.

**Important**: the CA_DataUploaderLib now targets netcoreapp3.0 instead of netstandard. This is needed to support unloading of plugins.

**Other changes:**
- CommandHandler.AddCommand now returns a deregistration function, so that added commands can be removed
- fixed: CommandHandler.AddCommand failed when subscribing to the same command with different names
- help no longer displays command accepted messages
- executing the same command repeatedly no longer shows multiple times using the up arrow


CommandHandlerTesterLib/TestLongActionPlugin.cs
```diff
@@ -0,0 +1,52 @@
...
+            Task.Run(DoLongAction);
+            return true;
+        }
+
+        private async Task DoLongAction()
+        {
+            try
+            {
+                Console.WriteLine("Waiting for the next multiple of 10 value for IterationSensor");
+                var val = await WhenSensorValue("IterationSensor", v => v % 10 == 0, TimeSpan.FromSeconds(12));
```
freddyrios, line 33: @copenhagenatomics could not come up with a way to improve on the lambda v => v % 10 == 0

only thing that comes to mind, would be potentially using expressions, so that we use the name of the lambda argument as the sensor name, like this:

`WhenSensorValue(IterationSensor => IterationSensor % 10 == 0, TimeSpan.FromSeconds(12));

But not sure that is much better.

CommandHandlerTesterLib/TestLongActionPlugin.cs
```diff
@@ -0,0 +1,52 @@
...
+                Console.WriteLine($"IterationSensor = {val}. Waiting for value to increase by 3");
+                var vector = await When(e => e["IterationSensor"].Value >= val + 3, TimeSpan.FromSeconds(5)); // alt way of waiting / can target multiple sensor values too
```
freddyrios, line 35: @copenhagenatomics I could add a Seconds convenience method to the base class, so we can do:
`await When(e => e["IterationSensor"].Value >= val + 3, Seconds(5))`
also
`await Task.Delay(Seconds(4));`
And perhaps a Delay method, so: `await Delay(Seconds(4))`

CommandHandlerTesterLib/TestLongActionPlugin.cs
```diff
@@ -0,0 +1,52 @@
...
+{
+    public class TestLongActionPlugin : LoopControlPlugin
+    {
+        public TestLongActionPlugin(CommandHandler cmd) : base(cmd)
+        {
+            AddCommand(nameof(LongAction), LongAction);
+            AddCommand("help", HelpMenu);
+        }
+
+        private bool HelpMenu(List<string> _)
```
freddyrios, line 16: could do in another PR, but plan to change it so that if I implement a plugin command, I always have help. Potentially the base class would be LoopControlCommand with a command name and description. It could also help take out the boilerplate around the action, as we could take some of the boilerplate away from the action, so that we get login automatically and potentially a separate action that runs on timeouts (instead of explicitely having to catch the TaskCanceledException).

copenhagenatomics: Please do.

CommandHandlerTesterLib/TestLongActionPlugin.cs
copenhagenatomics: not better

CommandHandlerTesterLib/TestLongActionPlugin.cs
copenhagenatomics: I like this: await Delay(Seconds(4))

copenhagenatomics (approved):
Super v1

CommandHandlerTesterLib/TestLongActionPlugin.cs
freddyrios: @copenhagenatomics then plan to keep it as is. Even with v1, plugin authors have the option of using local functions, wonder if that helps
```
bool IsMultipleOf10(double val) => val % 10 == 0;
var val = await WhenSensorValue("IterationSensor", IsMultipleOf10, Seconds(12));
```

CA_DataUploaderLib/CommandHandler.cs
```diff
@@ -55,28 +71,116 @@ public bool AssertArgs(List<string> args, int minimumLen)
...
+                    LoadPlugin(assembly);
+
+                if (isAuto)
+                    TrackPluginChanges("*.dll");
+                return true;
+            }
+
+            try
+            {
+                LoadPlugin(args[1]);
```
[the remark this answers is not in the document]
freddyrios: @copenhagenatomics what do you think, any value of keeping these extra options?

CommandHandlerTesterLib/TestLongActionPlugin.cs
freddyrios: @copenhagenatomics added PRs with this

CA_DataUploaderLib/LoopControlPlugin.cs
```diff
@@ -0,0 +1,81 @@
...
+        protected async Task<NewVectorReceivedArgs> When(Predicate<NewVectorReceivedArgs> condition, TimeSpan timeout)
+        {
+            using var cts = new CancellationTokenSource(timeout);
+            return await When(condition, cts.Token);
+        }
+
+        protected Task<NewVectorReceivedArgs> When(Predicate<NewVectorReceivedArgs> condition, CancellationToken token)
+        { 
+            var tcs = new TaskCompletionSource<NewVectorReceivedArgs>();
+            cmd.NewVectorReceived += OnNewValue;
```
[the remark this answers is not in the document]
freddyrios: already in new PR

CA_DataUploaderLib/CommandHandler.cs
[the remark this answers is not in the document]
freddyrios: removed in the latest PR.

CA_DataUploaderLib/CommandHandler.cs
```diff
@@ -13,14 +14,19 @@ public class CommandHandler : IDisposable
...
-        private CALogLevel _logLevel = IOconfFile.GetOutputLevel();
+        private readonly StringBuilder inputCommand = new StringBuilder();
+        private readonly Dictionary<string, List<Func<List<string>, bool>>> _commands = new Dictionary<string, List<Func<List<string>, bool>>>();
+        private readonly CALogLevel _logLevel = IOconfFile.GetOutputLevel();
         private readonly List<string> AcceptedCommands = new List<string>();
         private int AcceptedCommandsIndex = -1;
+        private readonly Dictionary<string, (AssemblyLoadContext ctx, IEnumerable<LoopControlPlugin> instances)> _runningPlugins = 
+            new Dictionary<string, (AssemblyLoadContext ctx, IEnumerable<LoopControlPlugin> instances)>();
+        private FileSystemWatcher _pluginChangesWatcher;
+        readonly Dictionary<string, object> _postponedPluginChangeLocks = new Dictionary<string, object>();
```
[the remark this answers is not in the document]
freddyrios: moved in the latest PR

CommandHandlerTesterLib/TestLongActionPlugin.cs
freddyrios: Implemented in latest PR

PR #57, opened 2021-02-01
enabled rpi temp by default

freddyrios:
- fixed recurrent empty lines printed by the RpiTemp code
- as its now on by default, its now possible to explicitely disable Rpi Temp: RpiTemp;RpiTemp;Disabled
- only add rpi temp to the vector when its enabled
- rpi temp now records both cpu and gpu


freddyrios:
@copenhagenatomics  it seems running commands in parallel, seems to be broken in .net core. Here is somewhat facing similar [.net core processes issue](https://github.com/dotnet/runtime/issues/33297#issuecomment-773915576).

It crashes the main process, with an exception that gets in the output but does not hit our exception handler in Program.Main.

The parallel commands are the new temperature ones + the RpiVersion ones that run on startup.

copenhagenatomics (approved):

Issue #58, opened 2021-02-08
issues running commands


freddyrios:
It seems running commands in parallel, seems to be broken in .net core. Here is somewhat facing similar [.net core processes issue](https://github.com/dotnet/runtime/issues/33297#issuecomment-773915576).

It crashes the main process, with an exception that gets in the output but does not hit our exception handler in Program.Main.

The parallel commands are the new temperature ones + the RpiVersion ones that run on startup.

_Originally posted by @freddyrios in https://github.com/copenhagenatomics/CA_DataUploader/issues/57#issuecomment-773927370_

freddyrios:
Besides the parallel commands issue on auto start, we also ran into the below after running for a few days / potentially related to running 2 commands every 100ms. 

@copenhagenatomics I will change this to run every 10 seconds, which is 100 times less frequent. Still concerned about it, because if whatever underlying resource is not cleaned on its own, there is a risk for this to trigger after 200-300 days.

PR #59, opened 2021-02-09
fix for #58 issues running commands

freddyrios:
The main fix that was confirmed to address the issue is we avoid running processes in parallel + ensure there are 10+ ms from the exit of one to the start of the next one.

Also did the following, which was confirmed to still reproduce the issue without the main fix:
- fixed missing Dispose of the Process, in an attempt to address a crash after 3 days running 2 child processes every 100 ms.
- removed previous timing based workaround, that could still hit the issue depending on the configuration of the system i.e. depends on the start time which in turns depends on the overall system complexity
- aligned process start with msdn docs removing possibility of deadlock. In one case if the standard error buffer got full and the child process would wait while the parent was waiting for the output to finish. In the other case in SerialNumberMapper if standard output buffer got full and the child process would wait while the parent waited for the child to exit.
- removed sudo usages that should not have been there

### DIFF
--- a/CA_DataUploaderLib/BaseSensorBox.cs
+++ b/CA_DataUploaderLib/BaseSensorBox.cs
@@ -73,12 +73,11 @@ namespace CA_DataUploaderLib
 
             while (_running)
             {
-                Thread.Sleep(100); //boards typically write a line every 100 ms (note we also way on exception cases)
-
                 try
                 {
                     ReadSensors();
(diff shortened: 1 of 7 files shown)


freddyrios:
Note that while this fixes the issue in a more reliable way, it is still a work around and we should keep track of https://github.com/dotnet/runtime/issues/33297.

copenhagenatomics (approved):

PR #63, opened 2021-02-19
.net 5 regression: child parallel processes crash

freddyrios:
The upgrade to net5.0 broke the workaround used in #59. However, the reason the original issue still triggered in some systems without the delay between processes was because the conflict also involved a process started by a third party API being used on the affected systems. So the delay is what in reality was avoiding the conflict with that process and the lock was not doing anything for that.

Due to this, a better workaround that is in practice a permament fix for our case is for the program that runs in the other systems to be adjusted so that the offending third party code runs at a time where we can be certain no other child processes are being ran by us.

This PR removes the earlier attempted workarounds, so that child processes can now again run in parallel without any kind of locks or delays. It also includes the unrelated change of _power to _current we have on the graphs.

### DIFF
--- a/CA_DataUploader/CA_DataUploader.csproj
+++ b/CA_DataUploader/CA_DataUploader.csproj
@@ -13,4 +13,11 @@
     <Content Include="USBmapper.sh;IO.conf" CopyToOutputDirectory="Always" ExcludeFromSingleFile="true" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="MinVer" Version="2.4.0">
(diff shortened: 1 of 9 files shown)


freddyrios:
As in the original, this is a workaround to https://github.com/dotnet/runtime/issues/33297. If it gets addressed we should move away from the workaround.

freddyrios:
Updated the description based on the findings after running a trace as suggested by microsoft. More info on https://github.com/dotnet/runtime/issues/33297#issuecomment-784063579.

copenhagenatomics (approved):

PR #69, opened 2021-03-09
Update build.yml

freddyrios:


freddyrios:
I had to change the checkout to pull the whole repository to get MinVer working again. This is because the checkout action would otherwise not pull the tags.

See https://github.com/actions/checkout/issues/448

PR #73, opened 2021-03-24
fixed: missing error info when failing to update a single plugin

freddyrios:


freddyrios:
@copenhagenatomics merging this + #72 to keep moving, but feel free to send any review comments

PR #75, opened 2021-04-09
minor changes to IO.Conf + improved scales initialization

freddyrios:
- moved some of the IOConf files code into the base class
- removed the unused IOconfTank
- improved the scales detection, similar to how oxygen sensor detection is done with the right baud rate when configured
- scales no longer require indicating a port number, like implied by its format
- improved reuse of sensor box initialization code


freddyrios:
@copenhagenatomics going through the code base preparing for the new architecture, so sending some intermediate steps as a smaller PR

freddyrios:
@copenhagenatomics going ahead and merging this to continue on next steps. Let me know if there is any feedback.

PR #76, opened 2021-04-14
New architecture

freddyrios:
- added exit confirmation + log extra info on exit. Reasons: 
  - some different users have stopped the program unintentionally.
  - it was not obvious whether the exit sequence was user initiated or another subsystem triggered it
- CommandHandler avoids repeating success msgs
- removed unused check on BaseSensorBox.ShowQueue.
- removed unused debugging only ovenhistory / use the command history instead
- removed some unused Dispose boilerplate and sealed a few classes to do it

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -149,7 +149,7 @@ publish/
 # Publish Web Output
 *.[Pp]ublish.xml
 *.azurePubxml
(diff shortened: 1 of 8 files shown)


freddyrios:
@copenhagenatomics merging this one too in order to continue moving forward on the loop side

Issue #62, opened 2021-02-17
Remove SerialNumberMapper.PortsChanged


freddyrios:
- It only works on Windows and it is the only reason we have the nuget dependency to System.Management.
- It is not used on by the CA_DataUploader and elsewhere is only used for informational purposes.

freddyrios:
@copenhagenatomics thoughts?

freddyrios:
discussed on meeting, we will remove it

PR #79, opened 2021-04-21
#62 removed SerialNumberMapper.PortsChanged

freddyrios:
* #62 removed SerialNumberMapper.PortsChanged
* fixed broken rpi temp sensors with recent changes
* cleanup: ServerUploader
  * the upload loop now better attempts to keep the 1 second frequency
  * introduced a couple private classes handling the plot connection and the related signing
  * removed the unused: UploadSensorMatch, PrintMyPlots and ListMyPlots
  * made some of the methods shorter and added more reuse in some cases


copenhagenatomics (approved):
OK

PR #80, opened 2021-04-23
introduced/moved board related settings to the IOconf classes

freddyrios:
Reasons:
- hardcoding these in BaseSensorBox required too much code to be unnecesarily overwritten in derived classes
- benefits cases where BaseSensorBox is not used i.e. VFD now uses "SkipBoardAutoDetection" that should speed up startup
- its now much easier to support boards that use different lines formats (see IOconfScale for an example)
- the IOconf classes now owns the decision to stop the system when losing communications to a that type of sensor
- MCUBoard now skips extra header lines during initialization, instead of only on SafeReopen / useful for the flow box

note: only the MCUBoard and BaseSensorBox (and derived classes) are using the new configurations at the moment.

Other improvements:
- VFD board comm initialization now skips trying to read serial while still checking a connection is made at the target baud rate.
- If an usb port is expected to have a given baud rate, we still attempt auto discovery with other baud rates and we print/log a message with the board info with the unexpected baud rate.
- We now show the attempted baud rate on failure messages connecting to a board
- Slowed down Server.Uploader on recurrent failures
- BaseSensorBox.CheckFails avoids unnecesary lists allocations on every loop cycle / now it does it only when we lost the connection to a board
- the connection to ignored sensors due to lost communication now is disposed, which could allow connecting to those for diagnostics while the system still runs

### DIFF
--- a/CA_DataUploaderLib/BaseSensorBox.cs
+++ b/CA_DataUploaderLib/BaseSensorBox.cs
@@ -20,9 +20,9 @@ namespace CA_DataUploaderLib
         protected CommandHandler _cmd;
         protected List<SensorSample> _values = new List<SensorSample>();
         protected List<MCUBoard> _boards = new List<MCUBoard>();
-        protected int expectedHeaderLines = 8;
         private string commandHelp;
         private bool disposed;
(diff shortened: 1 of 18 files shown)


copenhagenatomics (commented):
Super.

PR #83, opened 2021-05-25
fixed: connection retries on startup

freddyrios:
Reason: missed moving some code to use await that meant it throws AggregateException instead of the expected HttpRequestException


freddyrios:
@copenhagenatomics FYI merging this small fix.

PR #84, opened 2021-05-25
fixed: trying to use the last switchboard port logs recurring exceptions

freddyrios:
Reason: missed updating the port for the switchboard temperature so it clashes with the switchboard state of the last port.


freddyrios:
FYI @copenhagenatomics merging this one as its preventing use on systems with 4 heaters

PR #86, opened 2021-06-22
minor improvements

freddyrios:
- added always on switchboard actions (pass DateTime.MaxValue to the constructor). Usage of this should be limited to devices where the safe state is on (which is discouraged, as the safe state of devices should be off).
- SwitchboardAction.ToVectorSamples now reports the port is off if the time off has passed
- SwitchboardAction.GetRemainingOnSeconds now returns how many seconds since the time to turn off has passed as a negative number (it used to return 0 all the time)
- added SwitchBoardAction.Repeat that can be used to request the vector based action to be repeated


freddyrios:
@copenhagenatomics merging this one as is, will link it from the related PR that prompted the changes. Let me know if there are any comments.

PR #90, opened 2021-07-13
improved board state + less shutdowns + better sensor independence + better reconnects

freddyrios:
Breaking Change: 
- plugins must now take into account the system will not auto shutdown when losing temperature/heater boards. So in applications where a minimum temperature must be mantained the plugin can set alerts with emergency shutdown with the minimum operating temperature. If no emergency shutdown is desired, the plugin can track the temperature directly and do corresponding actions. When doing so, the temperature boards connection state must be taken into account in order to react appropiately when losing a connection board (which in the absence of temperature/heater redundancy would cause the temperature to drop).

Changes:
- loop now sends an alert instead of stopping the program on recurrent errors reaching a box i.e. could not recover connection to a board. This  happens if 5+ mins pass without being able to reach the box (100 attempts every 3 seconds). After this, it always continues trying to reconnect but approx. 4 times x hour (96 x day).
- oven decision logic now treats disconnected temperature boards like we do the 10k errors, which means it postpones making decision changes for up to 2 seconds, after which it turns off the heater until we get valid reads again.
- fixed unexpected extra switching time regression by avoiding reads blocking writes to the same board (which meant waiting for the next current print cycle to be done before acting). Note switching time is still delayed due to vector based decision, but that much is expected by design.
- recurrent reconnects now report the board states 0 (disconnected) and 1 (connecting). For this the receiving data (fully connected) state is now 6, while the initial connection gives the state 2 (connected).
- improved sensor independence across boards of the same type (unrelated read loops)
- rpi temperatures read loops no longer affect the thermocouples read loops
- more controlled reconnect attempts that avoids writing/reading while the reconnects and excessive messages caused by it.
- safe valves conf can now return all the board states names used for the sensor so that decision logic can detect the sensor boxes disconnects
- reduced amount of messages in the display when we lose connection to a box and attempt to reconnect
- fixed: states report boards one off no data reads states every few mins (which can unnecesarily affect decision logic). The fix also avoids the up to 2 second delay the existing implementation had in some cases before reporting no data is coming back. Also added early detection of hard disconnects in cases where the API was explicit about it, entering the improved reconnect logic right away.
- improved some of the errors messages by moving the board info to the end (making the relevant text more visible)
- improved incorrectly formatted lines detection, which avoids unnecesary exceptions being logged when broken lines are returned by temp hubs on reconnects

### DIFF
--- a/CA_DataUploader/Program.cs
+++ b/CA_DataUploader/Program.cs
@@ -30,12 +30,13 @@ namespace CA_DataUploader
                     CALog.LogInfoAndConsoleLn(LogID.A, "Now connected to server...");
 
                     int i = 0;
+                    var uploadThrottle = new TimeThrottle(100);
                     while (cmd.IsRunning)
                     {
                         var (sensorsSamples, vectorTime) = cmd.GetFullSystemVectorValues();
                         cloud.SendVector(sensorsSamples.Select(v => v.Value).ToList(), vectorTime);
                         Console.Write($"\r data points uploaded: {i++}"); // we don't want this in the log file. 
-                        cloud.Wait(100);
+                        uploadThrottle.Wait();
                         if (i == 20) DULutil.OpenUrl(cloud.GetPlotUrl());
                     }
                 }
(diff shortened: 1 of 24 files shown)


freddyrios:
@copenhagenatomics needed the board state + allow parallel read/write to stabilize the vector based changes. Rest came as part of stabilizing the latest version.

Will merge to prepare a proper release to test on more systems. Will address any review input on a separate PR.

PR #101, opened 2021-08-03
its now possible to expand the IO.conf file from outside this repository

freddyrios:


freddyrios:
@copenhagenatomics going forward and merging this for now, will be using it for the new board

PR #103, opened 2021-08-04
removed valves from data uploader IO.conf implementation

freddyrios:
reason: there is nothing supporting valves in the data uploader lib. Any consuming adding valve support must use the IOconfFileLoader to add and fully control their valve configurations.


freddyrios:
@copenhagenatomics FYI

PR #106, opened 2021-08-06
alerts no longer trigger on temporary thermocouples disconnects / 10k values

freddyrios:
Note: a warning is still sent if a thermocouple has been disconnected for 5+ minutes.

Additionally:
- added support for using other commands than emergencyshutdown when an alert fires. To run multiple commands use | as a separator i.e. `Alert;MyName;Sensorx<123;oven off | pump 0`
- fixed: temperature board disconnect detection in heating control is being ignored

### DIFF
--- a/CA_DataUploaderLib/Alerts.cs
+++ b/CA_DataUploaderLib/Alerts.cs
@@ -109,7 +109,10 @@ namespace CA_DataUploaderLib
             message = timestamp + message;
             logger.LogError(message);
             if (a.Command != default)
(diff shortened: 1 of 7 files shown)


freddyrios:
@copenhagenatomics in [this commit](https://github.com/copenhagenatomics/CA_DataUploader/pull/106/commits/6bba68ede1a73333333cc4a0fdeec4ad4c8da0cd) is the fix I mentioned earlier.

PR #119, opened 2021-09-07
removed logging of the removed switchboard state

freddyrios:
the switchboard state was removed in #115 but was still used for logging purposes. This finishes the removal of it.


freddyrios:
@Gnistrup FYI

PR #120, opened 2021-09-13
allow to add register byte based protocols to MCUBoard

freddyrios:
- it is now possible to add custom protocols detection to MCUBoard / including byte based protocols. For now it has the overhead of converting the data to a string of numbers, mainly to limit the already large changes (ideally the read data should go directly into the samples in the BaseSensorBox).
- read loop should now react to changes as they come, instead of having explicit pauses
- added IOconfMap.OnBoardDetected that can allow related config classes to react to the detected board type.
- added MCUBOard.TryReadAsciiLine public that helps implement custom ascii based protocol detectors
- removed IOConfOxygen, IOConfScale and related detection in MCUBoard as there is no functionality for it in the CA_DataUploader
- the MCUBoard class no longer inherits the SerialPort and uses composition instead. This reinforces the controls it does to avoid other operations running against the port while it reconnects + allows the move to a pipe reader which simplifies the higher complexity code for byte based protocols.
- fixed IO.conf in the CA_DataUploader to have ports starting at 1


freddyrios:
merging ahead of review to start preparing related PR. A separate PR will be created to address any review comments.

Gnistrup:
It does not seems to solve the problem
https://www.copenhagenatomics.com/plots/temperatureplot.php?DkgcvVISsS

PR #124, opened 2021-09-22
fixed memory leak introduced by #120

freddyrios:


CA_DataUploaderLib/MCUBoard.cs
```diff
@@ -410,13 +414,17 @@ private async Task<string> ReadLine()
...
                         throw new TimeoutException("timed out (soft)");
                     var buffer = res.Buffer;
-                    var readLine = TryReadLine(ref buffer, out var line);
-                    pipeReader.AdvanceTo(buffer.Start, buffer.End); // Tell the PipeReader how much of the buffer has been consumed.
+                    var readLine = tryReadLine(ref buffer, out var line);
+                    //if we don't have a full line we set the whole buffer  as examined so the next ReadAsync call waits for more data.
+                    //if we instead got a full line we avoid setting any data of the next line as examined,
+                    //as otherwise the next call to ReadLine will wait for yet more data to arrive causing a memory leak (buffer keeps growing) 
+                    var examinedPos = readLine ? buffer.Start : buffer.End;
+                    pipeReader.AdvanceTo(buffer.Start, examinedPos);
```
Gnistrup, line 422: I cant see how this will reduce the memory leak. If false, i.e. no line nothing is released from pipeline. Offcourse it will prevent throwing lines that is not yet fully received.

freddyrios: we just read a full line:
`pipeReader.AdvanceTo(buffer.Start, buffer.Start);`
we did not read a full line:
`pipeReader.AdvanceTo(buffer.Start, buffer.End);`

in the later case we really need more data, so setting examined to buffer.End does just that. Next call to ReadAsync will wait until we get more data or the cancelation token times out.

In the earlier case we did not need more data, so setting examined to buffer.Start makes it so the next call to ReadAsync returns inmediately so we can process the already available data right away. That was the leak, waiting for more lines in cases where we already had one. In a system when we hit that case once, we are likely to keep hitting it over time, so the amount of lines kept in the buffer kept growing. I added an unit test that confirms that last bit.

CA_DataUploaderLib/MCUBoard.cs
```diff
@@ -410,13 +414,17 @@ private async Task<string> ReadLine()
                     if (res.IsCanceled)
                         throw new TimeoutException("timed out (soft)");
                     var buffer = res.Buffer;
-                    var readLine = TryReadLine(ref buffer, out var line);
-                    pipeReader.AdvanceTo(buffer.Start, buffer.End); // Tell the PipeReader how much of the buffer has been consumed.
+                    var readLine = tryReadLine(ref buffer, out var line);
```
freddyrios, line 417: If tryReadLine throws the pipeReader is not advanced / this could potentially cause a tight loop if some custom read line parser throws exceptions. Note that parsers throwing exceptions must be avoided anyway as it prevents making progress on the data, nevertheless the tight loop would make the bad situation worse.

https://docs.microsoft.com/en-us/dotnet/standard/io/pipelines
there is a reference ReadSingleMessageAsync that shows an example avoiding the tight loop

Gnistrup (approved):

freddyrios:
seems to still leak: https://github.com/copenhagenatomics/CA_DataUploader/pull/120#issuecomment-925749609

PR #125, opened 2021-09-23
Prevent LoopControl stall from VFD not responding 

matias-cphatomics:
During the initial detection of boards connected to the system the system stall because the VFD never sends a line back. To overcome this CancellationTokens are passed where the library function .ReadAsync is used. 

The program catches an OperationCanceledException and returns values that indicate it was not able to receive anything from that port.

A UnitTest "ProcessingEmptyLineDoesNotStallInitialization" is created to ensure this does not stall the program for more than 4 seconds.

### DIFF
--- a/CA_DataUploaderLib/MCUBoard.cs
+++ b/CA_DataUploaderLib/MCUBoard.cs
@@ -279,7 +279,7 @@ namespace CA_DataUploaderLib
 
             // A cancellation token is made here rather than a simple timer since the ReadAsync function can hang
             // if a device never sends a line for it to read.
-            int millisecondsTimeout = 3000;
+            int millisecondsTimeout = 5000;
             using var cts = new CancellationTokenSource(millisecondsTimeout);
             var token = cts.Token;
 
@@ -391,7 +391,7 @@ namespace CA_DataUploaderLib
 
             // A cancellation token is made here rather than a simple timer since the ReadAsync function can hang
             // if a device never sends a line for it to read.
-            int millisecondsTimeout = 3000;
+            int millisecondsTimeout = 5000;
             using var cts = new CancellationTokenSource(millisecondsTimeout);
             var token = cts.Token;
             foreach (var detector in customProtocolDetectors)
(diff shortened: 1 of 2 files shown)


CA_DataUploaderLib/MCUBoard.cs
```diff
@@ -373,28 +389,43 @@ private bool TryReadOptionalCalibration(ref ReadOnlySequence<byte> buffer, out s
...
+
+                        // set all data as examined as we need more data to finish detection
+                        pipeReader.AdvanceTo(buffer.Start, buffer.End);
+
+                        if (result.IsCompleted)
+                            return default; //should only happen if the connection was closed
+                    } 
+                    catch (OperationCanceledException ex)
+                    {
+                        if (ex.CancellationToken == token)
```
freddyrios, line 421: if the token does not match, we should still throw;

CA_DataUploaderLib/MCUBoard.cs
```diff
@@ -373,28 +389,43 @@ private bool TryReadOptionalCalibration(ref ReadOnlySequence<byte> buffer, out s
...
+                        {
+                            CALog.LogColor(LogID.A, ConsoleColor.Red, $"Unable to read from {portName} ({baudRate}): " + ex.Message);
```
freddyrios, line 423: we should probably skip this logging, because at this step we are tyring to detect a third party device and the result was not detected.

it will get logged anyway when we try to read from it as a normal board

CA_DataUploaderLib/MCUBoard.cs
```diff
@@ -273,16 +273,27 @@ private async Task<bool> ReadSerialNumber()
...
             port.WriteLine("Serial");
-            var stop = DateTime.Now.AddSeconds(5);
             bool sentSerialCommandTwice = false;
             bool ableToRead = false;
             var finishedReadingHeader = false;
-            while (!finishedReadingHeader && DateTime.Now < stop)
+
+            // A cancellation token is made here rather than a simple timer since the ReadAsync function can hang
+            // if a device never sends a line for it to read.
+            int millisecondsTimeout = 3000;
```
freddyrios, line 282: to avoid extra testing needed I suggest to stick to 5 seconds here, so that we are not changing the max time allowed to read the header during initialization (we had some old boards that were slow to initialize and respond with a header).

freddyrios (approved):

CA_DataUploaderLib/MCUBoard.cs
matias-cphatomics: Alright, will just make it break resulting in return default.

PR #126, opened 2021-09-27
fix serialport memoryleak by updating to .net 6

freddyrios:
After #124 a second leak was found. Dump analysis showed it being in SerialPort and already fixed in dotnet 6. See https://github.com/dotnet/runtime/issues/55146#issuecomment-927344824.

This PR updates to .net 6. Note that:

- this means the repository now requires vs 2022 and additionally to enable preview builds via Tools -> Options -> Environment -> Preview Features -> Use previews of the .NET SDK
- CA.LoopControlPluginBase was left untouched still targetting .net 5, so that can continue using the same plugins during the transition. Most plugins don't use features that had breaking changes in the move from .net 5 to 6 and initial testing was positive, so it makes more sense to wait until we have a reason to do a breaking change for plugins.
- initial testing shows very clearly that the managed memory leak is gone. However, early testing still showed an increase of native memory, although at a much slower rate i.e. 3 MB in 5 hours. That rate is still problematic as it would imply above 400MB x month, although it slowed down to approx 120MB x month. Additionally another system explored running an older version also showed a memory growth pattern that stabilized later in the run.

### DIFF
--- a/CA_DataUploaderLib/ServerUploader.cs
+++ b/CA_DataUploaderLib/ServerUploader.cs
@@ -368,7 +368,7 @@ namespace CA_DataUploaderLib
 
             public byte[] GetPublicKey() => _rsaWriter.ExportCspBlob(false);
 
-            public byte[] GetSignature(byte[] data) => _rsaWriter.SignData(data, new SHA1CryptoServiceProvider());
+            public byte[] GetSignature(byte[] data) => _rsaWriter.SignData(data, SHA1.Create());
         }
     }
 }
--- a/.github/workflows/main-build.yml
+++ b/.github/workflows/main-build.yml
@@ -17,7 +17,7 @@ jobs:
       with:
         fetch-depth: 0
     - name: Setup .NET
-      uses: actions/setup-dotnet@v1
+      uses: actions/setup-dotnet@v1.8.2
       with:
         dotnet-version: 6.0.x
     - name: Restore dependencies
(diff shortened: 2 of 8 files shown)


CA_DataUploaderLib/ServerUploader.cs
```diff
@@ -368,7 +368,7 @@ public Signing(string loopName)
 
             public byte[] GetPublicKey() => _rsaWriter.ExportCspBlob(false);
 
-            public byte[] GetSignature(byte[] data) => _rsaWriter.SignData(data, new SHA1CryptoServiceProvider());
+            public byte[] GetSignature(byte[] data) => _rsaWriter.SignData(data, SHA1.Create());
```
matias-cphatomics, line 371: Is this change simply related to the dotnet update? It looks like the to functions have similar functionality.

freddyrios: yes, its part of the breaking changes. They have flagged that obsoleted and ask to use that base class's Create method. I don't know when they introduced that base class or if it was always there

Gnistrup (commented):
With this memory leaj we must update to 6.0 but I am not a big fan since it is a RC.

freddyrios:
@Gnistrup agree, but at least it will be releasing soon (november) and is under a go live license.

by the way, .net 5 was not under long term support, so[ it will be out of support by may next year](https://devblogs.microsoft.com/dotnet/net-core-releases-and-support/#net-core-versions-and-support) due to the .net 6 release in november.

PR #127, opened 2021-09-30
accepted user commands and alerts are now registered as events

freddyrios:
By registering these low frequency events it will be possible later to check relevant actions and alerts that have been triggered for the system

Also:
* fixed bug in multi command alerts
* CommandHandler.AlertFired has been removed, use CommandHandler.EventFired and check type == EventType.Alert instead


freddyrios:
created the PR too fast, adding a test for `EventFiredArgs.ToByteArray`

freddyrios:
merging this as is for now, will address any review comments on a separate PR.

PR #132, opened 2021-10-13
updates to make the config to plots values more predictable + allow system to start in more cases

freddyrios:
* system now reports the same amount of values for missing boards
* turned some local warnings into alerts that will also show in the event log
* system still starts when there are missing heaters
* system now reports device detection events, which adds device detection and calibration updates info to the event log
* system temperature names now include the loop name as a prefix
* the subsystems are now explicitely run via CommandHandler.RunSubSystems
   Reason: fixes bug where alerts related to missing boards were not being sent during initialization. This is because the alerts would trigger before the ServerUploader was subscribed to CommandHandler.EventFired


freddyrios:
@lt-copenhagenatomics FYI

PR #149, opened 2022-03-03
fixes #136

### LINKED ISSUE #136: prevent multiple extra spaces in commands from being detected as extra arguments
Specially plugins should not need to account for extra spaces in the commands sent to them.

In particular a command like this one turns into extra empty arguments, due to the 3 spaces at the end of the line: myplugin arg1 arg&nbsp;&nbsp;&nbsp;\n

freddyrios:
note that this only fixes issues where the extra spaces are all at the beginning or the end of the command line. It does *not* address a command with extra spaces within the arguments, such like this one:  oven 100     200 150

It is of course possible to solve, for example, by using a simple regex for the split. However, at that point we might as well consider using an existing solution to parse commands, such like https://github.com/dotnet/command-line-api

### DIFF
--- a/CA_DataUploaderLib/CommandHandler.cs
+++ b/CA_DataUploaderLib/CommandHandler.cs
@@ -167,7 +167,7 @@ namespace CA_DataUploaderLib
                 return;
             }
 
-            var cmd = cmdString.Split(' ').Select(x => x.Trim()).ToList();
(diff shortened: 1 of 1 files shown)


freddyrios:
@thomasjamatomics FYI addressed the specific scenarios in #136 for now, but kept it short and left some extra scenarios unsolved (see the description).

PR #154, opened 2022-03-23
fixed unclear message

freddyrios:
- added the connection state to the message about stopping switchboard actions, so that it states NodeUnreachable in cases where the whole node is detected as such (instead of communication problems with the board)
- the stopping/resuming switchboard actions message now only show in the B.log instead of in the screen for the operator. While it signals the point in time where actions actually resume, it was normally redundant as other messages from the read loop are usually shown (attempting to reconnect / reconnected). In this way, there is less noise in the screen without removing the ability of doing postmortem analysis with both the log entries. Operators can still look at the connection state in the plots without looking at the logs.


freddyrios:
FYI @ThomasHSkov

PR #155, opened 2022-03-25
uploads additional mcu board detected information

freddyrios:
Also the existing system change notification now displays calibration changes at the board level so its clear which board was being updated + if there is any calibration returned by the board its always printed.


freddyrios:
@lasse-tarp-CA testing + git sha are pending

lasse-tarp-CA (approved):

PR #157, opened 2022-04-05
removed unused features incompatible with multi node deployments

freddyrios:
- StopWhenLosingSensor: an emergency shutdown done by this unused board setting should be done instead by the decision logic and potentially by an explicitely configured alarm
- HandleSaltLeakage + its auto emergency shutdown + error info: this unused behavior should instead be done by an explicitely configured alarm + anyway needs to be aligned with any relevant hardware for leak detection


CA_DataUploaderLib/BaseSensorBox.cs
```diff
@@ -390,26 +384,6 @@ private SensorSample[] GetSamples(MCUBoard board)
             return samples;
         }
 
-        private void HandleSaltLeakage(MCUBoard board, SensorSample sensor)
-        {
-            if (sensor.GetType() == typeof(IOconfSaltLeakage))
-            {
-                if (sensor.Value < 3000 && sensor.Value > 0)  // Salt leakage algorithm.
```
freddyrios, line 397: @thomasjamatomics any objections to this removal? as far as I know this style of salt leakage detection configuration is not in use anywhere + can be done by combinations of math + alarms and/or introducing something more closely aligned with any hardware + multi node deployments

freddyrios: discussed offline, will remove

PR #163, opened 2022-04-26
small logging / events noise improvements

freddyrios:
* fixed: noise on startup for subsystems without (detected) boards

* removed log message that is noise for custom hosts

* phasing out help for local read commands / use the plots instead

* version and up now include the node name in their output (useful for distributed deployments)

* bad packages message is now logged as one event instead of multiple ones


freddyrios:
@lasse-tarp-CA name of the nodename boards property [here](https://github.com/copenhagenatomics/CA_DataUploader/pull/163/files#diff-1997ec9a64d88f655979827ca2632b877edcfba025dd8443aebe6a145ce9b6edR38).

PR #171, opened 2022-07-11
states in code exploration

freddyrios:


freddyrios:
Closing this as we are done with exploration.

We dropped the custom dsl approach in favor of scxml. For now we have a private generator that produces something similar to the c# code in this exploration but expands to handle enough of the scxml to be usable.

PR #172, opened 2022-07-20
features which allow machine generating xxxDataVector from IO.conf

thomasjamatomics:
Let's add these new features to the master branch.


CA.LoopControlPluginBase/CA.LoopControlPluginBase.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net5.0</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
```
freddyrios, line 4: let's remove this bit to postpone breaking changes to the current plugins

CA_DataUploaderLib/DataVector.cs
```diff
@@ -6,8 +6,11 @@ namespace CA_DataUploaderLib
 {
     public class DataVector
     {
-        public DateTime timestamp;
-        public List<double> vector;
+        public readonly DateTime timestamp;
+        public readonly List<double> vector;
+        public readonly VectorDescription _vectorDescription;
```
freddyrios, line 11: this field is unused

CA_DataUploaderLib/IOconf/IOConfFileLoader.cs
```diff
@@ -29,21 +29,23 @@ public static class IOconfFileLoader
             ("Node", (r, l) => new IOconfNode(r, l)),
         };
 
-        public static (string, IEnumerable<IOconfRow>) Load(List<IOconfRow> target)
+        public static (List<string>, IEnumerable<IOconfRow>) Load()
         {
             if (!File.Exists("IO.conf"))
-                throw new Exception($"Could not find the file {Directory.GetCurrentDirectory()}\\IO.conf");
+            {
+                throw new Exception($"Could not find the file {Directory.GetCurrentDirectory() + "\\IO.conf"}");
```
freddyrios, line 36: these are equivalent (the earlier is the idomatic version for string interpolation):
{Directory.GetCurrentDirectory()}\\\\IO.conf
{Directory.GetCurrentDirectory() + "\\\\IO.conf"

CA_DataUploaderLib/IOconf/IOconfRow.cs
```diff
@@ -12,12 +12,14 @@ public IOconfRow(string row, int lineNum, string type)
             Type = type;
             LineNumber = lineNum;
             var list = ToList();
-            if (list[0] != Type) throw new Exception("IOconfRow: wrong format: " + Row);
+            if (list[0] != Type && type != "Unknown") throw new Exception("IOconfRow: wrong format: " + Row);
```
freddyrios, line 15: If we go with this, we should probably still reject running with these somewhere else.

reason: it turns mistypes into unknown rows without letting the user know.

ps. without the specific type loaders we don't know how many values they will expand to anyway

CA_DataUploaderLib/VectorDescription.cs
```diff
@@ -7,19 +7,21 @@ namespace CA_DataUploaderLib
     [Serializable]
     public class VectorDescription
     {
-        public List<VectorDescriptionItem> _items;  // only public because we need to Serialize
+        public List<VectorDescriptionItem> _items;  // only public because we need to Serialize  -> Do NOT add or remove anything from this list.
```
freddyrios, line 10: its not impossible, but we do need to be **very** careful when doing it if we do a breaking change at some point

freddyrios (changes requested):

CA_DataUploaderLib/IOconf/IOconfRow.cs
```diff
@@ -12,12 +12,14 @@ public IOconfRow(string row, int lineNum, string type)
...
+            Name = list[1]; // could be overwritten elsewhere.
```
freddyrios, line 16: suggest to use this field in the UniqueKey + the field should be changed in the constructor of both IOconfOven and IOconfMap as in both cases list[1] is not the name.

PR #173, opened 2022-07-22
Now you can get and set values directly on DataVector

thomasjamatomics:
Since this is not tied to a specific IO.conf you can use it when generating state machines. And still use the names. However be aware that it may return null. (which suck)


freddyrios:
abandoning this PR as we decided to stick to having the consumers map the fields they use to field indexes based on the vector description and later on stick 100% to only using the indexes to get the values directly without any string lookups being involved.

PR #176, opened 2022-08-18
added base class for decision plugins

freddyrios:
Important: this is work in progress and none of the integration into the DataUploaderLib is done at the moment


freddyrios:
The reason this build is failing is because it requires:
- to publish a new major version of CA.LoopControlPluginBase (could be done a non release tag)
- update the reference to the above

PR #214, opened 2023-01-20
the uploader now only reports 3+ messages duplicates

freddyrios:
(before it was 2+ x minute)


freddyrios:
will retarget 3.x first

PR #215, opened 2023-01-20
the uploader now only reports duplicate errors for 3+ duplicates x minute

freddyrios:
before it used to be 2 x minute


freddyrios:
wrong PR target

PR #225, opened 2023-03-09
apply shared mprovements to AsyncReaderWriterLock

freddyrios:
Instead of 
```
await lock.AcquireReaderLock(token);
try
{
    // code goes here ... the lock is taken until the finally runs
}
finally
{
    lock.ReleaseReaderLock();
}
```
We now do:
```
using var _ = lock.AcquireReaderLock(cancellationToken);
// code goes here ... the lock is taken until the end of the current method
```

thanks to @cajuncoding sharing it here https://stackoverflow.com/a/64757462/66372


cajuncoding:
👍

PR #249, opened 2023-06-22
#576 BaseSensorBox: added immediate reconnect command

LarsWithCA:
Command: “reconnect *boxname*”. 
TaskCompletionSources are used to signal (to the ReconnectBoard-method) that reconnect should be attempted for a board with a specific name.

So far only tested running LoopControl on a PC. Will also need to test on a multipi system.

### DIFF
--- a/CA_DataUploaderLib/BaseSensorBox.cs
+++ b/CA_DataUploaderLib/BaseSensorBox.cs
@@ -646,7 +646,7 @@ namespace CA_DataUploaderLib
         private void LogData(MCUBoard board, string message) => CALog.LogData(LogID.B, $"{message} - {Title} - {board.ToShortDescription()}");
         private void LogInfo(MCUBoard board, string message) => CALog.LogInfoAndConsoleLn(LogID.B, $"{message} - {Title} - {board.ToShortDescription()}");
 
-        protected class AllBoardsState : IEnumerable<(string sensorName, ConnectionState State)>
+        public class AllBoardsState : IEnumerable<(string sensorName, ConnectionState State)>
         {
             private readonly ConnectionState[] _states;
             private readonly string[] _sensorNames;
--- a/CA_DataUploaderLib/BaseSensorBox.cs
+++ b/CA_DataUploaderLib/BaseSensorBox.cs
@@ -121,9 +121,12 @@ namespace CA_DataUploaderLib
                     a => a.Count == 2 && a[1] == map.BoxName,
                     a =>
                     {
-                        var currentTCS = reconnectTasks[map.BoxName];
-                        reconnectTasks[map.BoxName] = new TaskCompletionSource();
-                        currentTCS.TrySetResult(); //Completes the task signaling to ReconnectBoard to try again (if currently waiting)
+                        lock (reconnectTasks)
+                        {
+                            var currentTCS = reconnectTasks[map.BoxName];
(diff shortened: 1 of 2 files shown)


CA_DataUploaderLib/BaseSensorBox.cs
```diff
@@ -93,10 +95,37 @@ static void RegisterCustomBoardCommand(IOconfMap map, CommandHandler cmd, Dictio
...
+                {//if the board is not local we register the command validation with an empty action
+                    cmd.AddMultinodeCommand("reconnect", a => a.Count == 2 && a[1] == map.BoxName, _ => { });
+                    return;
+                }
+
+                if (map.McuBoard == null)
+                {
+                    cmd.AddMultinodeCommand(
+                        "reconnect", a => a.Count == 2 && a[1] == map.BoxName,
+                        _ => { CALog.LogData(LogID.A, $"reconnect failed due to missing board: {map.BoxName}"); });
```
freddyrios, line 114: when we share the new feature, we need to explicitely state that this is only to reconnect boards that were detected/connected when the system was first started

CA_DataUploaderLib/BaseSensorBox.cs
```diff
@@ -93,10 +95,37 @@ static void RegisterCustomBoardCommand(IOconfMap map, CommandHandler cmd, Dictio
...
+
+                reconnectTasks[map.BoxName] = new TaskCompletionSource();
+                cmd.AddMultinodeCommand(
+                    "reconnect",
+                    a => a.Count == 2 && a[1] == map.BoxName,
+                    a =>
+                    {
+                        var currentTCS = reconnectTasks[map.BoxName];
+                        reconnectTasks[map.BoxName] = new TaskCompletionSource();
+                        currentTCS.SetResult(); //Completes the task signaling to ReconnectBoard to try again (if currently waiting)
```
freddyrios, line 126: not sure about the details, but as I understand it is usually best to stick to TrySetResult to signal the tcs

CA_DataUploaderLib/BaseSensorBox.cs
```diff
@@ -93,10 +95,37 @@ static void RegisterCustomBoardCommand(IOconfMap map, CommandHandler cmd, Dictio
...
+                }
+
+                reconnectTasks[map.BoxName] = new TaskCompletionSource();
+                cmd.AddMultinodeCommand(
+                    "reconnect",
+                    a => a.Count == 2 && a[1] == map.BoxName,
+                    a =>
+                    {
+                        var currentTCS = reconnectTasks[map.BoxName];
+                        reconnectTasks[map.BoxName] = new TaskCompletionSource();
```
freddyrios, line 125: We are writing here to a dictionary that we read from a different thread. Not if this could be or not an issue, as at least we are updating an existing entry in the dictionary

freddyrios (approved):

CA_DataUploaderLib/BaseSensorBox.cs
LarsWithCA: fixed

CA_DataUploaderLib/BaseSensorBox.cs
LarsWithCA: added locking around use of dictionary

LarsWithCA:
Tested and works in multipi setup.

PR #251, opened 2023-07-08
added support for GenericOutput

freddyrios:
This allows to tie any vector field to an output by specifyng a simple template in a config line.

This makes it possible to have subsystems that are done purely on plugins + config lines.

The generic aspect here means 'not specific to a board type'. However, without host customizations the support is limited to boards using the standard copenhagen atomics communications format (usb serial port with 115200 baud rate that returns csv lines at 10Hz + receives commands as a line of text).

To use, add a line to the IO.conf:

`GenericOutput;OutputName;BoxName;DefaultValue;command with $field`

Some examples:

- `GenericOutput;myheater;heaterbox;0;p1 $myheater_onoff 3`
- `GenericOutput;myheater;heaterbox;0;p1 on 3 ${myheater_onoff}00%`

The first 3 are values in the line are the standard type of the line, name of the sensor/actuator and box name.

The DefaultValue indicates the target that is used if no new vector data is received on the system. For example, if a node in a multinode host loses connection to the rest of the nodes.

The `command with $field` is the text line that will be sent to the box whenever there is a new value for the field. If there are no changes to the field, the system resends the command every 2 seconds, to re-assert the desired state of the connected box. The $field must be replaced with the name of the channel as shown in the plots, keeping the $ prefix.

In the first example, the board receives the command `p1 1 3`, eveyrtime the field myheater_onoff is set to 1 and whenever it changes to 0 it gets `p1 0 3`. Note these are not valid commands for the copenhagen atomics switchboard as it uses p1 on 3. Instead, the second example can be used which produces 'p1 on 3 100%' and 'p1 on 3 000%'.

### DIFF
--- a/UnitTests/IOConfFileLoaderTests.cs
+++ b/UnitTests/IOConfFileLoaderTests.cs
@@ -30,6 +30,35 @@ namespace UnitTests
             Assert.AreEqual(405, math.Calculate(new() { {"heater1", 400} }));
         }
 
+        [TestMethod]
+        public void CanLoadGenericOutputLine()
+        {
+            using var _ = TestableIOconfFile.Override(new() { GetMap = () => new[]{new IOconfMap("Map;acserial;realacbox2", 0)} });
+            var rowsEnum = IOconfFileLoader.ParseLines(new[] { "GenericOutput;generic_ac_on;realacbox2;0;p1 $heater1_onoff 3" });
+            var rows = rowsEnum.ToArray();
+            Assert.AreEqual(1, rows.Length);
+            Assert.IsInstanceOfType(rows[0], typeof(IOconfGenericOutput));
+            var output = (IOconfGenericOutput)rows[0];
+            Assert.AreEqual("generic_ac_on", output.Name);
+            Assert.AreEqual(0, output.DefaultValue);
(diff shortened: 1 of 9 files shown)


UnitTests/IOConfFileLoaderTests.cs
```diff
@@ -30,6 +30,35 @@ public void CanLoadMathLine()
...
+            var rowsEnum = IOconfFileLoader.ParseLines(new[] { "GenericOutput;generic_ac_on;realacbox2;0;p1 on 3 ${heater1_onoff}00%" });
+            var rows = rowsEnum.ToArray();
+            Assert.AreEqual(1, rows.Length);
+            Assert.IsInstanceOfType(rows[0], typeof(IOconfGenericOutput));
+            var output = (IOconfGenericOutput)rows[0];
+            Assert.AreEqual("generic_ac_on", output.Name);
+            Assert.AreEqual(0, output.DefaultValue);
+            Assert.AreEqual("heater1_onoff", output.TargetField);
+            Assert.AreEqual("p1 on 3 100%", output.GetCommand(1));
+        }
```
LarsWithCA, line 61: how about adding a test or two verifying that we get FormatExceptions when the CommandRegex is unsuccessful?

PR #259, opened 2023-09-12
added OvenProportionalControlUpdates

freddyrios:
This allows live updates to the proportional control arguments of the heating control. Note that the updated values are lost when restarting the system.

The configuration line requires specifying the maximum value for the different proportional control arguments. The meaning and format of the arguments is the same as in the Oven configuration lines.

The feature is only supported at the oven area level, so all the heaters in the area get the same values. When the arguments have not been changed, the system uses the configuration values specified in the Oven configuration lines for each heater.

* **IO.conf lines:** 
    * `OvenProportionalControlUpdates;[MaxProportionalGain];[MaxControlPeriod];[MaxOutputPercentage]` e.g. `OvenProportionalControlUpdates;10;00:01:00;100`
* **user commands:** 
    * set proportional gain for oven area 2: `ovenarea 2 pgain 0.5`
    * set control period for oven area 1: `ovenarea 1 controlperiodseconds 100 `
    * set max output for oven area 3: `ovenarea 3 maxoutput 70`
* **plugin fields:** `ovenarea{number}_pgain`, `ovenarea{number}_controlperiodseconds`, `ovenarea{number}_maxoutput`. The fields are state fields that retain their value once set. The change is only reflected in the next control period.

### DIFF
--- a/CA_DataUploaderLib/HeatingController.cs
+++ b/CA_DataUploaderLib/HeatingController.cs
@@ -19,9 +19,10 @@ namespace CA_DataUploaderLib
 
             var ovens = IOconfFile.GetOven().ToList();
             var allAreas = IOconfFile.GetOven().Select(x => x.OvenArea).Distinct();
+            var ovenProportionalControlUpdatesConf = IOconfFile.GetEntries<IOconfOvenProportionalControlUpdates>().SingleOrDefault();
 
             //notice that the decisions states and outputs are handled by the registered decisions, while the switchboard inputs and actuations are handled by the switchboard controller
-            cmd.AddDecisions(allAreas.Select(a => new OvenAreaDecision(new($"ovenarea{a}", a))).ToList());
+            cmd.AddDecisions(allAreas.Select(a => new OvenAreaDecision(new($"ovenarea{a}", a, ovenProportionalControlUpdatesConf))).ToList());
             cmd.AddDecisions(heatersConfigs.Select(h => new HeaterDecision(h, ovens.SingleOrDefault(x => x.HeatingElement.Name == h.Name))).ToList());
             SwitchBoardController.Initialize(cmd);
         }
@@ -117,6 +118,13 @@ namespace CA_DataUploaderLib
                 public double on { get => _latestVector[_indexes.on]; set => _latestVector[_indexes.on] = value; }
                 public double nextcontrolperiod { get => _latestVector[_indexes.nextcontrolperiod]; set => _latestVector[_indexes.nextcontrolperiod] = value; }
                 public double controlperiodtimeoff { get => _latestVector[_indexes.controlperiodtimeoff]; set => _latestVector[_indexes.controlperiodtimeoff] = value; }
+                public double pgain { get => _latestVector[_indexes.pgain]; set => _latestVector[_indexes.pgain] = value; }
+                public bool pgain_defined => _indexes.pgain != -1;
+                public double controlperiodseconds { get => _latestVector[_indexes.controlperiodseconds]; set => _latestVector[_indexes.controlperiodseconds] = value; }
+                public bool controlperiodseconds_defined => _indexes.controlperiodseconds != -1;
+                public double maxoutput { get => _latestVector[_indexes.maxoutput]; set => _latestVector[_indexes.maxoutput] = value; }
+                public bool maxoutput_defined => _indexes.maxoutput != -1;
+
                 public bool TemperatureBoardsConnected
                 {
                     get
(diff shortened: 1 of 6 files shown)


CA_DataUploaderLib/HeatingController.cs
```diff
@@ -366,6 +407,23 @@ internal void MakeDecision(Events e, double data = 0)
...
+                            case (States.Off or States.Running, Events.pgain):
+                                pgain = Math.Min(data, _config.OvenProportionalControlUpdatesConf!.MaxProportionalGain);
+                                break;
+                            case (States.Off or States.Running, Events.controlperiodseconds):
+                                controlperiodseconds = Math.Min(data, _config.OvenProportionalControlUpdatesConf!.MaxControlPeriod.TotalSeconds);
+                                break;
+                            case (States.Off or States.Running, Events.maxoutputs):
+                                maxoutput = Math.Min(data / 100, _config.OvenProportionalControlUpdatesConf!.MaxOutputPercentage);
+                                break;
+                            case (States.Off or States.Running, Events.vector):
```
LarsWithCA, line 419: is this case necessary?

CA_DataUploaderLib/IOconf/IOconfOvenProportionalControlUpdates.cs
```diff
@@ -0,0 +1,35 @@
...
+
+            (MaxProportionalGain, MaxControlPeriod, MaxOutputPercentage) = (maxpGain, maxPeriod, maxOutput / 100d);
+        }
+
+        public const string TypeName = "OvenProportionalControlUpdates";
+
+        /// <remarks><see cref="IOconfOven.ProportionalGain"/></remarks>
+        public double MaxProportionalGain { get; }
+        public TimeSpan MaxControlPeriod { get; }
+        public double MaxOutputPercentage { get; }
```
LarsWithCA, line 33: add a comment stating that this has a value between 0 and 1?

UnitTests/HeaterDecisionTests.cs
```diff
@@ -280,11 +303,83 @@ public void CanEmergencyShutdownWithoutOven()
...
+        {
+            var ovenProportionalControlUpdatesConf = new IOconfOvenProportionalControlUpdates("OvenProportionalControlUpdates;5;00:02:00;100", 0);
+            NewOvenAreaDecisionConfig(new("ovenarea0", 0, ovenProportionalControlUpdatesConf));
+            NewOvenAreaDecisionConfig(new("ovenarea1", 1, ovenProportionalControlUpdatesConf));
+            NewHeaterDecisionConfig(NewConfig.WithProportionalGain(2).Build());
+            CheckDecisionsBehaviorMatchesAProportionalGainOf2();
+        }
+
+        [TestMethod]
+        public void PControlUsesUsesMaxConfigurationForCommandValuesOutsideOfRange()
```
LarsWithCA, line 325: 2x Uses

UnitTests/HeaterDecisionTests.cs
```diff
@@ -280,11 +303,83 @@ public void CanEmergencyShutdownWithoutOven()
...
+        {
+            var ovenProportionalControlUpdatesConf = new IOconfOvenProportionalControlUpdates("OvenProportionalControlUpdates;2;00:00:30;100", 0);
+            NewOvenAreaDecisionConfig(new("ovenarea0", 0, ovenProportionalControlUpdatesConf));
+            NewOvenAreaDecisionConfig(new("ovenarea1", 1, ovenProportionalControlUpdatesConf));
+            NewHeaterDecisionConfig(NewConfig.WithProportionalGain(1).WithMaxOutput(0.5).WithControlPeriodSeconds(15).Build());
+            CheckDecisionsBehaviorMatchesAProportionalGainOf2(new[] { "ovenarea 0 pgain 10", "ovenarea 0 maxoutput 150", "ovenarea 0 controlperiodseconds 120" });
+        }
+
+        [TestMethod]
+        public void PControlUsesUsesMaxConfigurationForVectorValuesOutsideOfRange()
```
LarsWithCA, line 335: 2x Uses

LarsWithCA (approved):

CA_DataUploaderLib/HeatingController.cs
freddyrios: yes, without it a plugin can update the values outside the range and the heater decision would end up using those. Removing it breaks test:PControlUsesUsesMaxConfigurationForVectorValuesAboveAllowedRange

It does relies on the order of plugins, where the oven area decision runs before the heater decisions. Additionally built in decisions run before any plugin, so it is always the value within range that is used. 

Note I did not do it this way for the minimum valid values, so that the fix for that issue applies to both the new feature and when using values from the configuration file.

PR #261, opened 2023-10-12
654 loopcontrol handle new status hex value from boards

LarsWithCA:


CA_DataUploaderLib/IOconf/IOconfOut230Vac.cs
```diff
@@ -46,25 +46,25 @@ public SwitchBoardResponseParser (bool expectCommandConfirmations)
...
                 if (numbers.Count < 4)
-                    return null; //invalid line, must have at least 4 currents
+                    return (null, 0); //invalid line, must have at least 4 currents
                 if (numbers.Count == 4)
                     numbers.Add(10000);//some versions did not report temperature
-                return numbers;
+                return (numbers, status);
             }
 
             public override bool MatchesValuesFormat(string line) => _oldswitchBoxCurrentsRegex.IsMatch(line);
```
LarsWithCA, line 70: @freddyrios should this line be:
`public override bool MatchesValuesFormat(string line) => _oldswitchBoxCurrentsRegex.IsMatch(line) || base.MatchesValuesFormat(line);`
?

CA_DataUploaderLib/BaseSensorBox.cs
```diff
@@ -542,8 +547,8 @@ private async Task<(bool, string?)> TryReadLineWithStallDetection(MCUBoard board
             }
         }
 
-        /// <returns>the list of doubles, otherwise <c>null</c></returns>
-        protected virtual List<double>? TryParseAsDoubleList(MCUBoard board, string line) =>
+        /// <returns>the list of doubles (potentially <c>null</c>) and a status value</returns>
```
freddyrios, line 550: null = failure to parse the line as a double list e.g. we don't have a separate boolean like in the common TryParse pattern

CA_DataUploaderLib/IOconf/BoardSettings.cs
```diff
@@ -35,19 +36,25 @@ public void SetCalibrationAtIndex(string defaultCalibration, char value, int ind
...
-                if (!_hasCommaSeparatedNumbers.IsMatch(line))
-                    return null;
+                var status = 0U;
+                var lineMatch = _hasCommaSeparatedNumbers.Match(line);
+                if (!lineMatch.Success)
+                    return (null, status);
 
-                return line.Split(",".ToCharArray())
+                var statusString = lineMatch.Groups["Status"].Value.Trim(',', ' ');
+                if (!string.IsNullOrEmpty(statusString) && statusString.StartsWith("0x"))
```
freddyrios, line 50: we should use the regex to give you exactly the part of the string you want

something like 

`(?:,\s*0x(?<Status>[0-9a-fA-F]+))?`

CA_DataUploaderLib/IOconf/BoardSettings.cs
```diff
@@ -35,19 +36,25 @@ public void SetCalibrationAtIndex(string defaultCalibration, char value, int ind
...
+                    uint.TryParse(statusString[2..], NumberStyles.HexNumber, CultureInfo.InvariantCulture, out status);
+
+                return (line.Split(",".ToCharArray())
```
freddyrios, line 53: we should use the regex to get the line without the status

CA_DataUploaderLib/IOconf/IOconfOut230Vac.cs
freddyrios: yes, as that's the value lines we are allowing via TryParseAsDoubleList

freddyrios (approved):

CA_DataUploaderLib/IOconf/IOconfOut230Vac.cs
LarsWithCA: @freddyrios  how could it have worked up until now?

CA_DataUploaderLib/IOconf/BoardSettings.cs
LarsWithCA: done

CA_DataUploaderLib/BaseSensorBox.cs
LarsWithCA: I updated the text to say this.

CA_DataUploaderLib/IOconf/IOconfOut230Vac.cs
freddyrios: @LarsWithCA the method is only used in MCUBoard.SkipExtraHeaders where a failure would only result in unnecesarily skipping some value lines.

On initial connection it is not really used for this board due to the setting saying 8 expected header lines (we only skip above 8).

It only leaves reconnects and an extra delay to connect/get values is very likely to be ignored.

LarsWithCA: fixed

PR #263, opened 2023-10-27
Added RepeatMilliseconds to the GenericOutput configuration line

freddyrios:
* Added RepeatMilliseconds to the GenericOutput configuration line
* fixed vector based timeout in LastAction and added related unit test while also making those unit tests faster by using the TimeProvider for time passing in tests

The line format is now: `GenericOutput;OutputName;BoxName;DefaultValue;command with $field;[RepeatMilliseconds]`

So now these 2 are possible, where the second repeats the command every 2 minutes:
* `GenericOutput;myheater;heaterbox;0;p1 on 3 ${myheater_onoff}00%`
* `GenericOutput;myheater;heaterbox;0;p1 on 3 ${myheater_onoff}00%;120000`


freddyrios:
@LarsWithCA FYI

PR #268, opened 2023-11-22
added initial sensor redundancy support

freddyrios:
The redundancy support can be used to treat a group of sensors as if it was only one sensor for related control functions. To do this, the system uses the median of all the specified sensors. Sensors are skipped from the median calculation if connection to the sensor's box or node is lost.

IOconf lines:
* RedundantSensors;name;sensor1;sensor2;...;sensorn
* RedudantValidRange;name;min;max i.e. if a sensor gives values outside the range it is considered defective or disconnected and skipped from the median calculation
* RedundantInvalidDefault;name;invaliddefaultvalue i.e. specifies a value to use if we can't get data from any of the sensors; choose a value that would cause the related control function to go into its default state


CA_DataUploaderLib/Redundancy.cs
```diff
@@ -0,0 +1,233 @@
...
+        }
+
+        private class IOconfRedundantValidRange : IOconfRedundant
+        {
+            public (double min, double max) ValidRange { get; }
+
+            public IOconfRedundantValidRange(string row, int lineNum) : base(row, lineNum, "RedundantValidRange")
+            {
+                var vals = ToList();
+                if (vals.Count < 4) throw new FormatException($"Too little values. Format: RedundantValidRange;Name;MinValue;MaxValue. Row {Row}");
```
LarsWithCA, line 77: little->few

CA_DataUploaderLib/Redundancy.cs
```diff
@@ -0,0 +1,233 @@
...
+        }
+
+        private class IOconfRedundantInvalidDefault : IOconfRedundant
+        {
+            public double InvalidDefault { get; }
+
+            public IOconfRedundantInvalidDefault(string row, int lineNum) : base(row, lineNum, "RedundantInvalidDefault")
+            {
+                var vals = ToList();
+                if (vals.Count < 3) throw new FormatException($"Too little values. Format: RedundantInvalidDefault;Name;InvalidDefaultValue. Row {Row}");
```
LarsWithCA, line 92: little->few

CA_DataUploaderLib/Redundancy.cs
```diff
@@ -0,0 +1,233 @@
...
+                if (!vals[2].TryToDouble(out var invalidDefault))
+                    throw new FormatException($"Failed to parse invalid default value. Format: RedundantValidRange;Name;InvalidDefaultValue. Row {Row}");
```
LarsWithCA, line 94: RedundantValidRange ->RedundantInvalidDefault

CA_DataUploaderLib/Redundancy.cs
```diff
@@ -0,0 +1,233 @@
...
+        public Redundancy(CommandHandler cmd)
+        {
+            var configs = IOconfRedundant.ToDecisionConfigs(IOconfFile.GetEntries<IOconfRedundant>());
+            cmd.AddDecisions(configs.Select(r => new Decision(r)).ToList());
+            SwitchBoardController.Initialize(cmd);
+        }
+
+        public static void RegisterSystemExtensions()
+        {
+            IOconfFileLoader.AddLoader("RedundantSensors", (row, lineNumber) => new IOconfRedundantSensors(row, lineNumber));
```
LarsWithCA, line 21: please put into string properties in the relevant classes that can then be referenced :)

UnitTests/RedundancyDecisionTests.cs
```diff
@@ -0,0 +1,93 @@
...
+            Assert.AreEqual(14.0, Field("red5"));
+            Field("c") = 17;
+            Field("cbox") = (int)BaseSensorBox.ConnectionState.NoDataAvailable;
+            MakeDecisions();
+            Assert.AreEqual(16.5, Field("red5"), "must ignore stale sensor data and use median of the remaining 4");
+            Field("cbox") = (int)BaseSensorBox.ConnectionState.Connected;
+            Field("c") = 3000;
+            MakeDecisions();
+            Assert.AreEqual(16.5, Field("red5"), "must ignore stale sensor data outside the range and use median of the remaining 4");
+            Field("abox") = Field("bbox") = Field("cbox") = Field("dbox") = Field("ebox") = (int)BaseSensorBox.ConnectionState.NoDataAvailable;
```
LarsWithCA, line 64: suggestion: add before this line, to make sure the sensor is included again if it produces a valid value:
Field("c") = 13;
MakeDecisions();
Assert.AreEqual(14.0, Field("red5"));

CA_DataUploaderLib/Redundancy.cs
```diff
@@ -0,0 +1,233 @@
+﻿using CA.LoopControlPluginBase;
+using System.Collections.Generic;
+using System;
+using CA_DataUploaderLib.IOconf;
+using System.Linq;
+using CA_DataUploaderLib.Extensions;
```
LarsWithCA, line 6: nit: sort usings

LarsWithCA (approved):

PR #275, opened 2024-01-17
Implemented name validation on all IOconfRows

LarsWithCA:
Exceptions: Account, Nodes, Map-lines, IOconfControlWebUI, Oven+OvenProportionalControlUpdates (IOconfDriver). 
LoopName has a special validation (allowing hyphen). 
Also making sure that no rows use the same name (ignoring case).#257

Examples of error messages:

image

image

### DIFF
--- a/CA_DataUploaderLib/IOconf/IOconfMap.cs
+++ b/CA_DataUploaderLib/IOconf/IOconfMap.cs
@@ -119,5 +119,7 @@ namespace CA_DataUploaderLib.IOconf
         };
 
         public override string ToString() => $"{BoxName} - {USBPort ?? SerialNumber}";
+
+        protected override void ValidateName(string name) { } // no validation
     }
 }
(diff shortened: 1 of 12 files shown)


CA_DataUploaderLib/IOconf/IOconfMap.cs
```diff
@@ -119,5 +119,7 @@ public BoardSettings BoardSettings
         };
 
         public override string ToString() => $"{BoxName} - {USBPort ?? SerialNumber}";
+
+        protected override void ValidateName(string name) { } // no validation
```
freddyrios, line 123: do we have some other code in IOconfMap validating the box name format? the state field can be used in plugins, so it would be good if this also is forced to the valid range

freddyrios (approved):

CA_DataUploaderLib/IOconf/IOconfMap.cs
LarsWithCA: you're right. This needs validation. The tricky bit is that serial/port is initially set as the name (which is likely not a valid name), and then overwritten with the real name...

LarsWithCA: done

freddyrios (approved):

PR #277, opened 2024-01-24
CommandHandler: Implemented async version of Console.ReadKey

LarsWithCA:
So the program doesn't hang when getting shut down. 
Made Stop()-method public.

### DIFF
--- a/CA_DataUploaderLib/CommandHandler.cs
+++ b/CA_DataUploaderLib/CommandHandler.cs
@@ -66,8 +66,8 @@ namespace CA_DataUploaderLib
             _commandRunner = runner ?? new DefaultCommandRunner();
             _mapper = mapper;
             _fullsystemFilterAndMath = new Lazy<ExtendedVectorDescription>(GetFullSystemFilterAndMath);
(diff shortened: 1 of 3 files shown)


CA_DataUploaderLib/CommandHandler.cs
```diff
@@ -355,20 +357,20 @@ private void OnUserCommandAccepted(string cmdString)
...
+                else if (info.Value.Key == ConsoleKey.DownArrow)
                     RestoreNextAcceptedCommand();
                 else
                 {
-                    Console.Write(info.KeyChar);
-                    inputCommand.Append(info.KeyChar);
+                    Console.Write(info.Value.KeyChar);
+                    inputCommand.Append(info.Value.KeyChar);
                 }
                 info = Console.ReadKey(true);
```
freddyrios, line 375: remember to update other users of ReadKey

freddyrios (approved):

CA_DataUploaderLib/CommandHandler.cs
LarsWithCA: done

PR #281, opened 2024-02-15
fixed: MathVectorExpansion is not thread safe

freddyrios:
The problem with MathVectorExpansion not being thread safe, is that it made calls to CommandHandler.MakeDecision* methods not thread safe.

There were some thread offs going this way to solve the MakeDecision* thread safety:
- pro: consumers no longer need to be concerned about the thread safety of this method
- cons: extra overhead by using an ObjectPool to reuse dictionaries in MathVectorExpansion and an allocation of an NCalc Expression in IOConfMath.Calculate (this is a small object though, as the compiled expression is reused).


freddyrios:
@LarsWithCA FYI

PR #282, opened 2024-02-15
changes to slow detection

freddyrios:
- precise time of repeat events is not reported (so repeated slow cluster events within 1 minute get merged by existing duplicate messages logic)
- slow cluster is detected/reported when it takes at least 1 second to make decisions
- slow vector/event uploads are reported when it takes at least 15 seconds in an upload request


freddyrios:
@lasse-tarp-CA FYI

PR #284, opened 2024-03-13
223 Extend LoopControlDecision to expose list of fields referenced by the plugin

LarsWithCA:


CA_DataUploaderLib/CommandHandler.cs
```diff
@@ -55,6 +55,7 @@ public ChannelReader<DataVector> GetReceivedVectorsReader()
         public bool IsRunning => !_exitCts.IsCancellationRequested;
         public CancellationToken StopToken => _exitCts.Token;
         public Task RunningTask => _runningTaskTcs.Task;
+        public List<LoopControlDecision> Decisions => _decisions;
```
freddyrios, line 58: Consider either returning a copy or something that can't be changed

LarsWithCA: changed to be a IReadOnlyList

PR #287, opened 2024-04-05
ServerUploader: detecting invalid values (NaN/inf) in the vector and sending error as low frequency event (once every 5 minutes). #801

LarsWithCA:
Example IO.conf:
image

B.log:
image

Events:
image

### DIFF
--- a/CA_DataUploaderLib/ServerUploader.cs
+++ b/CA_DataUploaderLib/ServerUploader.cs
@@ -120,19 +120,19 @@ namespace CA_DataUploaderLib
                 var fullVectorData = fullVector.Data;
                 var uploadData = new double[uploadDesc.Length];
                 int j = 0;
-                StringBuilder invalidValues = new();
+                StringBuilder? invalidValues = null;
                 for (int i = 0; i < fullVector.Data.Length; i++)
                 {
                     if (!uploadMap[i]) continue;
 
                     if (double.IsNaN(fullVectorData[i]))
-                        invalidValues.Append($", {uploadDesc._items[j].Descriptor}=NaN");
+                        (invalidValues ??= new()).Append($", {uploadDesc._items[j].Descriptor}=NaN");
                     if (double.IsInfinity(fullVectorData[i]))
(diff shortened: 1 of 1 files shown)


CA_DataUploaderLib/ServerUploader.cs
```diff
@@ -94,19 +109,30 @@ private void SendVector(DataVector vector)
...
+            if (!string.IsNullOrWhiteSpace(invalidValueMessage))
+                SendLowFrequencyEvent(new EventFiredArgs("Warning - invalid values detected: " + invalidValueMessage[2..], EventType.Log, DateTime.UtcNow));
 
-            DataVector WithOnlyUploadFields(DataVector fullVector)
+            DataVector FilterOnlyUploadFieldsAndCheckInvalidValues(DataVector fullVector, out string invalidValueMessage)
             {
                 var fullVectorData = fullVector.Data;
                 var uploadData = new double[uploadDesc.Length];
                 int j = 0;
+                StringBuilder invalidValues = new();
```
freddyrios, line 123: This allocates memory on every vector even when not tripping

CA_DataUploaderLib/ServerUploader.cs
```diff
@@ -37,6 +37,8 @@ public sealed class ServerUploader : ISubsystemWithVectorData
         private readonly IIOconf _ioconf;
         private readonly CommandHandler _cmd;
         private readonly TaskCompletionSource<PlotConnection> _connectionEstablishedSource = new();
+        private readonly Stopwatch _timeSinceLastLowFrequencyEvent = new();
```
freddyrios, line 40: This watch should probably be specific to the special double values detection.

freddyrios (approved):

CA_DataUploaderLib/ServerUploader.cs
LarsWithCA: fixed

CA_DataUploaderLib/ServerUploader.cs

PR #289, opened 2024-04-16
fixes persistently broken connection to the server

freddyrios:
Context: changes to connection pools in .net 6 introduced a bug where a connection can remain permanently broken in the connection pool and all requests get stuck due to it. Here is an issue where one such case that gets a connection stuck is mentioned: https://github.com/dotnet/runtime/issues/81989

Also:
* adds retries to retrieving plots ids as well as covers more exception cases under the retries (some which may only happen with the new settings).
* fixes failure to refresh the IP if the server IP changes, by ensuring a new connection is used after 15 minutes


freddyrios:
abandoning due to wrong target branch

PR #290, opened 2024-04-16
fixes persistently broken connection to the server

freddyrios:
Context: changes to connection pools in .net 6 introduced a bug where a connection can remain permanently broken in the connection pool and all requests get stuck due to it. Here is an issue where one such case that gets a connection stuck is mentioned: https://github.com/dotnet/runtime/issues/81989

Also:
* adds retries to retrieving plots ids as well as covers more exception cases under the retries (some which may only happen with the new settings).
* fixes failure to refresh the IP if the server IP changes, by ensuring a new connection is used after 15 minutes

### DIFF
--- a/CA_DataUploaderLib/ServerUploader.cs
+++ b/CA_DataUploaderLib/ServerUploader.cs
@@ -8,6 +8,7 @@ using System.Diagnostics;
 using System.IO;
 using System.IO.Compression;
 using System.Linq;
+using System.Net;
 using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Net.Http.Json;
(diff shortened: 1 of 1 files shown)


CA_DataUploaderLib/ServerUploader.cs
```diff
@@ -558,6 +559,38 @@ public static async Task<PlotConnection> Establish(string loopName, byte[] publi
                 }
             }
 
+            private static async Task<(int plotId, string plotName)> GetPlotIDAsyncWithRetries(string loopName, HttpClient client, string loginToken, byte[] publicKey, byte[] signedVectorDescription, CancellationToken cancellationToken)
+            {
+                int failureCount = 0;
```
LarsWithCA, line 564: where is failureCount  incremented?

CA_DataUploaderLib/ServerUploader.cs
```diff
@@ -558,6 +559,38 @@ public static async Task<PlotConnection> Establish(string loopName, byte[] publi
                 }
             }
 
+            private static async Task<(int plotId, string plotName)> GetPlotIDAsyncWithRetries(string loopName, HttpClient client, string loginToken, byte[] publicKey, byte[] signedVectorDescription, CancellationToken cancellationToken)
```
LarsWithCA, line 562: Should we also add this to master? https://github.com/copenhagenatomics/CA_LoopCode/issues/790

freddyrios: Yes, I would say the only bit of the PR that I would not merge is the change to ConnectTimeout **and** the catching of Exception for retries. On the later I also read they changed exceptions for HttpClient so it all goes to to the same exceptions type with a reason

CA_DataUploaderLib/ServerUploader.cs
freddyrios: No where 🙃. Will fix.

LarsWithCA: It's also an issue in GetLoginTokenWithRetries

LarsWithCA (approved):

PR #292, opened 2024-04-17
Upgrade DataUploader to .NET8 #457

LarsWithCA:
https://github.com/copenhagenatomics/CA_LoopCode/issues/457

### DIFF
--- a/CA_DataUploaderLib/BaseSensorBox.cs
+++ b/CA_DataUploaderLib/BaseSensorBox.cs
@@ -22,15 +22,15 @@ namespace CA_DataUploaderLib
         private readonly CommandHandler _cmd;
         protected readonly List<SensorSample.InputBased> _values;
         protected readonly List<SensorSample.InputBased> _localValues;
-        private readonly (IOconfMap map, SensorSample.InputBased[] values, int boardStateIndexInFullVector)[] _boards = Array.Empty<(IOconfMap map, SensorSample.InputBased[] values, int boardStateIndexInFullVector)>();
-        protected readonly AllBoardsState _boardsState = new(Enumerable.Empty<IOconfMap>());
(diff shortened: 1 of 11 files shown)


CA_DataUploaderLib/BaseSensorBox.cs
```diff
@@ -22,15 +22,15 @@ public class BaseSensorBox : ISubsystemWithVectorData
         private readonly CommandHandler _cmd;
         protected readonly List<SensorSample.InputBased> _values;
         protected readonly List<SensorSample.InputBased> _localValues;
-        private readonly (IOconfMap map, SensorSample.InputBased[] values, int boardStateIndexInFullVector)[] _boards = Array.Empty<(IOconfMap map, SensorSample.InputBased[] values, int boardStateIndexInFullVector)>();
-        protected readonly AllBoardsState _boardsState = new(Enumerable.Empty<IOconfMap>());
-        private readonly Dictionary<MCUBoard, SensorSample.InputBased[]> _boardSamplesLookup = new();
+        private readonly (IOconfMap map, SensorSample.InputBased[] values, int boardStateIndexInFullVector)[] _boards = [];
```
freddyrios, line 25: Is the performance characteristics of this equivalent to the Array.Empty? E.g. not worried about this specific use, just in general for future use if this allocates more

freddyrios (approved):

CA_DataUploaderLib/BaseSensorBox.cs
LarsWithCA: https://github.com/dotnet/docs/blob/main/docs/csharp/language-reference/operators/collection-expressions.md#conversions

> The compiler uses static analysis to determine the most performant way to create the collection declared with a collection expression. For example, the empty collection expression, [], can be realized as xref:System.Array.Empty%60%601?displayProperty=nameWithType if the target won't be modified after initialization.

PR #298, opened 2024-04-23
Fixed RedundancySensors not working with expanded fields.

LarsWithCA:


CA_DataUploaderLib/Redundancy.cs
```diff
@@ -50,20 +50,27 @@ private class IOconfRedundantSensors : IOconfRedundant
...
             {
                 Sensors = ToList().Skip(2).ToList();
-                SensorBoardStates = Sensors.Select(s => new List<string>()).ToList();
                 if (Sensors.Count < 1)
                     throw new FormatException($"Missing sensors. Format: {RowType};Name;Sensor1;Sensor2...;Sensorn. Line: {Row}");
             }
 
             public override void ValidateDependencies(IIOconf ioconf)
             {
+                foreach (var sensor in Sensors)
```
freddyrios, line 63: note that we also validate these are in the vector

freddyrios (approved):

PR #300, opened 2024-04-29
ServerUploader: added logging.

LarsWithCA:


CA_DataUploaderLib/ServerUploader.cs
```diff
@@ -640,18 +654,20 @@ static async Task<(int plotId, string plotName)> GetPlotIDAsync((string loopName
...
                         return (result.StringBefore(" ").ToInt(), result.StringAfter(" "));
                     }
-                    catch(Exception ex)
+                    catch (Exception ex)
                     {
                         var contentTask = response?.Content?.ReadAsStringAsync(cancellationToken);
                         var error = contentTask != null ? await contentTask : null;
                         if (!string.IsNullOrEmpty(error))
-                            OnError($"Failure getting plot id: {error}");
+                            OnError($"Failure getting plot id: {error} {(ex is HttpRequestException rex && rex.HttpRequestError != HttpRequestError.Unknown ? "$(HttpStatusCode={rex.StatusCode}, HttpRequestError={rex.HttpRequestError})" : "")}");
```
freddyrios, line 665: I think we can keep the simpler version in this catch because it will get logged after the throw; at the end -> at RunWithRetries by the "Failed to ..., attempting to reconnect in 5 seconds." message

e.g. the log here is extra logging to capture only the response details that otherwise are hard to log on the higher catch

freddyrios (approved):

CA_DataUploaderLib/ServerUploader.cs

PR #304, opened 2024-05-07
logs lines when receiving partial board headers

freddyrios:
This is usually accompanied by a timeout.

Also:
- updated the expected message in these cases and no longer log the *expected* OperationCanceledException.
- the time out is now always logged, while before it dependended on timing (due to the soft exit with the while condition check)

### DIFF
--- a/CA_DataUploaderLib/MCUBoard.cs
+++ b/CA_DataUploaderLib/MCUBoard.cs
@@ -455,15 +455,23 @@ namespace CA_DataUploaderLib
             public string? Calibration { get; private set; }
             public string? UpdatedCalibration { get; private set; }
 
+            private readonly HeaderDepedendencies Dependencies;
+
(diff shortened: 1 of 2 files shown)


CA_DataUploaderLib/MCUBoard.cs
```diff
@@ -591,5 +603,16 @@ internal void CopyTo(MCUBoard mcuBoard)
                 mcuBoard.UpdatedCalibration = UpdatedCalibration;
             }
         }
+
+        public class HeaderDepedendencies(TimeProvider timeProvider, Action<LogID, string> logInfo, Action<LogID, string> logError, Action<LogID, string, Exception> logException)
```
LarsWithCA, line 607: typo:  HeaderDepedendencies -> HeaderDependencies

LarsWithCA (approved):

PR #315, opened 2024-05-29
Close connection to unmapped devices + enabled muting errors in ServerUploader.

LarsWithCA:


CA_DataUploaderLib/ServerUploader.cs
```diff
@@ -471,7 +477,7 @@ async Task<bool> Post()
 
             foreach (var e in events)
             {
-                if (TrackDuplicate(e.Data)) continue;
+                if (TrackDuplicate(e.Data) || (_muteErrors && e.Data.StartsWith("Failed sharing decision"))) continue;
```
freddyrios, line 480: The check should come from an extension point, instead of a hardcoded message check.

freddyrios (approved):

CA_DataUploaderLib/ServerUploader.cs

Issue #314, opened 2024-05-28
Some valid Math-lines fails validation


LarsWithCA:
example:

```
Math;PumpOverheat;20>600
Math;FloorLeak_Detected;20<600
Math;FailCondition;PumpOverheat||FloorLeak_Detected    <-- this bad boy
```

PR #316, opened 2024-05-30
IOconfMath: changing the way the list of parameters/sources are found

LarsWithCA:
#314

### DIFF
--- a/CA_DataUploaderLib/IOconf/IOconfMath.cs
+++ b/CA_DataUploaderLib/IOconf/IOconfMath.cs
@@ -93,9 +93,12 @@ namespace CA_DataUploaderLib.IOconf
                     expression.Accept(this);
             }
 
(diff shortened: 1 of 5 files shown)


freddyrios (approved):

CA_DataUploaderLib/IOconf/IOconfMath.cs
```diff
@@ -55,32 +55,47 @@ public double Calculate(Dictionary<string, object> values)
...
+            public override void Visit(Function function)
             {
-                sources.Add(name);
-                args.Result = 1;
-            };
+                foreach (var expression in function.Expressions)
+                    expression.Accept(this);
+            }
+
+            public override void Visit(LogicalExpression expression) { }
```
freddyrios, line 96: double check these do not lead to identifiers

PR #317, opened 2024-06-04
BaseSensorBox.ReadSensors: sending "Status" to the board in case the error bit of the status value is set.

LarsWithCA:
+ Implemented new class MultilineMessageReceiver for handling messages from the board (e.g. in response to "Status") spanning multiple lines. The lines are delimited by "Start of" and "End of" and logged as one message. While receiving a multiline message, no value lines can be received.


freddyrios (approved):
some type of event log rate limit should still apply to the multiline messages

PR #322, opened 2024-06-26
added Sustained filter type

freddyrios:
1. config line: `Filter; name; Sustained; FilterLength; condition` 
    e.g. `Filter; tempConsistentlyHighFor1Hour; Sustained; 3600; temp1 > 400`
2. Note it does **not** support hidesource
3. condition must be a boolean expression, such like: `sensor1 > 100`
4. condition supports the same syntax as Math config lines, as long as the expression returns true/false e.g. only `sensor1` is not allowed, but `sensor1 == 1` is allowed
5. the output is 1 if the condition returns true in every cycle during filter length seconds
6. the name must **not** be the same as any of its sources or any other name in the vector / channels in plot
7. the name of the filter is used as is e.g. "_filter" is not automatically added to the name
8. the sources can be any field/channel, including other filters and math. However, note that only input and regular filters will include the very latest value in the result, while other cases will only consider up to the previous cycle value.
9. the filter works well with very long filter lengths. However, like all system state, the filter starts fresh after a system restart
10. the filter will work correctly in the presence of leader changes (older filters have a queue that gets cleared on leader changes, but this filter keeps a few values in the vector state which gets shared all the time)

### DIFF
--- a/CA_DataUploaderLib/FilterVectorExpansion.cs
+++ b/CA_DataUploaderLib/FilterVectorExpansion.cs
@@ -10,7 +10,7 @@ namespace CA_DataUploaderLib;
 public class FilterVectorExpansion
 {
     private readonly List<IOconfFilter.DecisionFilter> _decisionFilters;
-    private readonly List<LegacyFilterSample> _legacyInputFilters;
+    private readonly List<LegacyFilter> _legacyInputFilters;
     private readonly CALogLevel _logLevel;
 
     /// <remarks>
--- a/CA_DataUploaderLib/FilterVectorExpansion.cs
+++ b/CA_DataUploaderLib/FilterVectorExpansion.cs
@@ -14,7 +14,7 @@ public class FilterVectorExpansion
     private readonly CALogLevel _logLevel;
 
     /// <remarks>
-    /// <paramref name="inputs"/> is updated by adding classic input filters and removing hidden sources.
+    /// <paramref name="inputs"/> is updated by adding legacy input filters and removing hidden sources.
     /// </remarks>
     public FilterVectorExpansion(List<VectorDescriptionItem> inputs, IEnumerable<IOconfFilter> configuredFilters, CALogLevel logLevel)
     {
@@ -67,7 +67,7 @@ public class FilterVectorExpansion
     }
 
     /// <remarks>
-    /// <paramref name="inputs"/> is updated by adding classic input filters and removing hidden sources.
+    /// <paramref name="inputs"/> is updated by adding legacy input filters and removing hidden sources.
     /// </remarks>
     public void ApplyLegacyFilters(List<SensorSample> inputs)
     {
--- a/CA_DataUploaderLib/FilterVectorExpansion.cs
+++ b/CA_DataUploaderLib/FilterVectorExpansion.cs
@@ -16,10 +16,10 @@ namespace CA_DataUploaderLib
         /// <remarks>
         /// <paramref name="inputs"/> is updated by adding classic input filters and removing hidden sources.
         /// </remarks>
-        public FilterVectorExpansion(List<VectorDescriptionItem> inputs, Func<IEnumerable<IOconfFilter>> getFilters, CALogLevel logLevel)
+        public FilterVectorExpansion(List<VectorDescriptionItem> inputs, IEnumerable<IOconfFilter> configuredFilters, CALogLevel logLevel)
         {
             _logLevel = logLevel;
-            var allfilters = getFilters().ToList();
+            var allfilters = configuredFilters.ToList();
             _decisionFilters = allfilters.Where(f => f.IsDecisionFilter).ToList();
             _legacyInputFilters = GetLegacyFilters();
             inputs.AddRange(_legacyInputFilters.Select(m => new VectorDescriptionItem("double", m.Output.Name, DataTypeEnum.Input)));
--- a/CA_DataUploaderLib/FilterVectorExpansion.cs
+++ b/CA_DataUploaderLib/FilterVectorExpansion.cs
@@ -1,0 +1,89 @@
+﻿#nullable enable
+using CA_DataUploaderLib.Helpers;
+using CA_DataUploaderLib.IOconf;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace CA_DataUploaderLib
+{
+    public class FilterVectorExpansion
+    {
+        private readonly List<IOconfFilter> _decisionFilters;
+        private readonly List<FilterSample> _legacyInputFilters;
+        private readonly CALogLevel _logLevel;
+
+        /// <remarks>
+        /// <paramref name="inputs"/> is updated by adding classic input filters and removing hidden sources.
+        /// </remarks>
+        public FilterVectorExpansion(List<VectorDescriptionItem> inputs, Func<IEnumerable<IOconfFilter>> getFilters, CALogLevel logLevel)
+        {
+            _logLevel = logLevel;
+            var allfilters = getFilters().ToList();
+            _decisionFilters = allfilters.Where(f => f.IsDecisionFilter).ToList();
+            _legacyInputFilters = GetLegacyFilters();
+            inputs.AddRange(_legacyInputFilters.Select(m => new VectorDescriptionItem("double", m.Output.Name, DataTypeEnum.Input)));
+            RemoveLegacyFiltersHiddenSources(inputs, i => i.Descriptor);
+
+            List<FilterSample> GetLegacyFilters()
+            {
+                var filters = allfilters.Where(f => !f.IsDecisionFilter).ToList();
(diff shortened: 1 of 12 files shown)


CA_DataUploaderLib/IOconf/IOconfFilter.cs
```diff
@@ -26,20 +33,96 @@ public IOconfFilter(string row, int lineNum) : base(row, lineNum, "Filter")
...
+        }
+
+        private static (LogicalExpression sustainedExpression, List<string> sourceNames) InitSustainedFilter(List<string> list, string row, string format)
+        {
+            try
+            {
+                var (sustainedExpression, sourceNames) = IOconfMath.CompileExpression(list[4]);
+                // Perform test calculation using default input values
+                if (IOconfMath.Calculate(sourceNames.ToDictionary(s => s, s => (object)0), sustainedExpression) is not bool)
+                    throw new FormatException($"Only boolean filter expressions are supported: {row}{Environment.NewLine}{format}");
```
LarsWithCA, line 65: suggestion: "Only boolean expressions are supported for Sustained filters: "

CA_DataUploaderLib/IOconf/IOconfFilter.cs
```diff
@@ -26,20 +33,96 @@ public IOconfFilter(string row, int lineNum) : base(row, lineNum, "Filter")
...
 
         public override IEnumerable<string> GetExpandedSensorNames(IIOconf ioconf)
         {
             yield return NameInVector;
         }
+
+        public IEnumerable<string> GetDecisionFields() => 
+            filterType == FilterType.Sustained 
+                ? [NameInVector, NameInVector + "_targettime"]
+                : throw new InvalidOperationException($"Unexpected call to GetDecisionFields for filter type {Row}");
```
LarsWithCA, line 87: "... for filter type {_filterType_}"?

CA_DataUploaderLib/IOconf/IOconfFilter.cs
```diff
@@ -26,20 +33,96 @@ public IOconfFilter(string row, int lineNum) : base(row, lineNum, "Filter")
...
+                throw new InvalidOperationException($"Unexpected Initialize call for non decision filter: {Row}");
+
+            _decisionFieldIndex = fields.IndexOf(Name);
+            if (_decisionFieldIndex < 0) throw new ArgumentException($"{Name} was not found in received vector fields", nameof(fields));
+            _sustainedTargetTimeFieldIndex = fields.IndexOf(Name + "_targettime");
+            if (_sustainedTargetTimeFieldIndex < 0) throw new ArgumentException($"{Name} was not found in received vector fields", nameof(fields));
+            foreach (var source in SourceNames)
+            {
+                if (!fields.Contains(source))
+                    throw new ArgumentException($"Filter {Name} uses {source} which was not found in received vector fields", nameof(fields));
```
LarsWithCA, line 124: in case there are more sources not found - how about listing them all in this message?

CA_DataUploaderLib/FilterVectorExpansion.cs
```diff
@@ -0,0 +1,89 @@
...
+                filter.Input(inputs);
+                inputs.Add(filter.Output);
+            }
+
+            RemoveLegacyFiltersHiddenSources(inputs, i => i.Name);
+        }
+
+        private void RemoveLegacyFiltersHiddenSources<T>(List<T> list, Func<T, string> getEntryName)
+        {
+            if (_logLevel == CALogLevel.Debug)
```
LarsWithCA, line 79: you could also check this at the call site - then we don't need to store the loglevel?

CA_DataUploaderLib/FilterVectorExpansion.cs
```diff
@@ -0,0 +1,89 @@
...
+    public class FilterVectorExpansion
+    {
+        private readonly List<IOconfFilter> _decisionFilters;
+        private readonly List<FilterSample> _legacyInputFilters;
+        private readonly CALogLevel _logLevel;
+
+        /// <remarks>
+        /// <paramref name="inputs"/> is updated by adding classic input filters and removing hidden sources.
+        /// </remarks>
+        public FilterVectorExpansion(List<VectorDescriptionItem> inputs, Func<IEnumerable<IOconfFilter>> getFilters, CALogLevel logLevel)
```
LarsWithCA, line 19: would you rather inject the getFilters Func - instead of just the list of filters?

CA_DataUploaderLib/IOconf/IOconfFilter.cs
freddyrios: It would be a bug, so I would rather have the whole row.  Changing it to this, but we can talk if that's still not ok: Unexpected call to GetDecisionFields in {Row}

CA_DataUploaderLib/FilterVectorExpansion.cs
```diff
@@ -0,0 +1,89 @@
...
+namespace CA_DataUploaderLib
+{
+    public class FilterVectorExpansion
+    {
+        private readonly List<IOconfFilter> _decisionFilters;
+        private readonly List<FilterSample> _legacyInputFilters;
+        private readonly CALogLevel _logLevel;
+
+        /// <remarks>
+        /// <paramref name="inputs"/> is updated by adding classic input filters and removing hidden sources.
```
LarsWithCA, line 17: align: classic/legacy

CA_DataUploaderLib/FilterVectorExpansion.cs
```diff
@@ -0,0 +1,89 @@
...
+                var filtersWithoutItem = filters
+                    .SelectMany(f => f.SourceNames.Select(s => new { Filter = f, Source = s }))
+                    .Where(f => !inputs.Any(i => i.Descriptor == f.Source))
+                    .ToList();
+                if (filtersWithoutItem.Count == 0)
+                    return filters.Select(x => new FilterSample(x)).ToList();
+
+                var names = string.Join(',', filtersWithoutItem.Select(f => f.Filter.Name).Distinct());
+                var sources = string.Join(',', filtersWithoutItem.Select(f => f.Source).Distinct());
+                throw new InvalidOperationException($"Misconfigured filters detected.{Environment.NewLine}Filters: {names}{Environment.NewLine}Missing sources: {sources}");
```
LarsWithCA, line 40: nit: Missing source(s)

CA_DataUploaderLib/MathVectorExpansion.cs
```diff
@@ -37,29 +38,47 @@ public void Initialize(IEnumerable<string> vectorFields)
...
+
+            private readonly Dictionary<int, string> _fieldsByIndex;
+            private readonly ObjectPool<Dictionary<string, object>> _reusableDictionaryPool;
+            public readonly DataVector Vector { get; }
+            private readonly Dictionary<string, object> _reusableFieldsDictionary;
+
+            public MathContext(Dictionary<int, string> fieldsByIndex, ObjectPool<Dictionary<string, object>> reusableDictionaryPool, DataVector vector)
             {
-                reusableDictionaryPool.Return(reusableFiedsDictionary);
+                _fieldsByIndex = fieldsByIndex ?? throw new InvalidOperationException("Initialize has not been called before calling Apply");
```
LarsWithCA, line 68: fieldsByIndex argument cannot be null, right?

UnitTests/FilterVectorExpansionTests.cs
```diff
@@ -0,0 +1,100 @@
...
+            ApplyFilterCycleAndAssertExpectedValues(expansion, mathExpansion, ref vector, 1, 6, 0);
+            ApplyFilterCycleAndAssertExpectedValues(expansion, mathExpansion, ref vector, 1, 5, 0);
+            ApplyFilterCycleAndAssertExpectedValues(expansion, mathExpansion, ref vector, 1, 7, 1);
+        }
+
+        [TestMethod]
+        public void AppliesSustainedFilterWithLessThanCondition()
+        {
+            var inputsDesc = ToDescItemInputs(["MyName"]);
+            var expansion = new FilterVectorExpansion(inputsDesc, GetFilters("Filter;MyFilter;Sustained;3;3 <= MyName - 2"), CALogLevel.Normal);
```
LarsWithCA, line 62: what about also testing out other boolean expressions, like: ==, !=, if (bla) true else false

LarsWithCA (approved):

CA_DataUploaderLib/Helpers/LegacyFilterSample.cs
```diff
@@ -6,44 +6,23 @@
 
 namespace CA_DataUploaderLib.Helpers
 {
-    public enum FilterType
+    public class LegacyFilterSample(string name, FilterType filterType, double length, List<string> sourceNames, bool hideSource)
```
LarsWithCA, line 9: what about just calling it LegacyFilter?

PR #324, opened 2024-07-10
backport board status support

freddyrios:


freddyrios:
abandoned because this should target 3.x

PR #333, opened 2024-09-02
Implemented global InvalidValueDelay used by all Redundant-lines as a possibility to hide/manage board data outages.

LarsWithCA:


CA_DataUploaderLib/Redundancy.cs
```diff
@@ -240,7 +266,14 @@ internal void MakeDecision()
...
 
                     validValues = validValues[..validValuesCount];
-                    value = !validValues.IsEmpty ? _config.Calculate(validValues) : _config.DefaultInvalidValue;
+                    if (!validValues.IsEmpty)
+                    {
+                        value = _config.Calculate(validValues);
+                        if (_config.InvalidValueDelay > 0)
+                            invalidValueCount = 0;
+                    }
+                    else if (_config.InvalidValueDelay == 0 || invalidValueCount++ >= _config.InvalidValueDelay)
```
LarsWithCA, line 275: @freddyrios in this case `invalidValueCount` will keep increasing if there are never any valid values - is that what we want?

freddyrios: I suggest to change it to not increase if you already reached the limit. 

An alternative to a count is using one of the time approaches, but given time to implement we could also keep it as is.

freddyrios: For reference I remembered where we have a time check, it's in the sustained filter: https://github.com/copenhagenatomics/CA_DataUploader/blob/master/CA_DataUploaderLib%2FIOconf%2FIOconfFilter.cs#L121-L134

In fact, a cumbersome workaround with the existing version would have been a combo of math and sustained. Sustained could be used to give you 1 if the filter goes into the invalid default for long enough and math that uses the new value if it is not the invalid default and otherwise the previous/invalid value depending if the filter is saying 1 or 0.

PR #334, opened 2024-09-06
Post updated calibration to boards

LarsWithCA:
https://github.com/copenhagenatomics/CA_LoopCode/issues/938

### DIFF
--- a/CA_DataUploaderLib/SystemChangeNotificationData.cs
+++ b/CA_DataUploaderLib/SystemChangeNotificationData.cs
@@ -39,7 +39,7 @@ namespace CA_DataUploaderLib
                     writer.WriteString("CompileDate", board.CompileDate);
                     writer.WriteString("GitSha", board.GitSha);
                     writer.WriteString("PcbVersion", board.PcbVersion);
-                    writer.WriteString("Calibration", board.Calibration);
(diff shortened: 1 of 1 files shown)


CA_DataUploaderLib/SystemChangeNotificationData.cs
```diff
@@ -39,7 +39,7 @@ public byte[] ToBoardsSerialInfoJsonUtf8Bytes(DateTime timeSpan)
                     writer.WriteString("CompileDate", board.CompileDate);
                     writer.WriteString("GitSha", board.GitSha);
                     writer.WriteString("PcbVersion", board.PcbVersion);
-                    writer.WriteString("Calibration", board.Calibration);
+                    writer.WriteString("Calibration", board.UpdatedCalibration); //Note: intentionally writing UpdatedCalibration value to Calibration property
```
freddyrios, line 42: Before merging double check we always set it vs. Only when it was updated

freddyrios (approved):

CA_DataUploaderLib/SystemChangeNotificationData.cs
LarsWithCA: When reading the calibration from the header, we always set `Calibration = UpdatedCalibration`:
https://github.com/copenhagenatomics/CA_DataUploader/blob/a86d85b22b0f30e5f49667d38963eaa69c568889/CA_DataUploaderLib/MCUBoard.cs#L543-L548

PR #335, opened 2024-09-13
ServerUploader: added method for getting plot name

LarsWithCA:


CA_DataUploaderLib/ServerUploader.cs
```diff
@@ -188,7 +199,11 @@ public async Task<HttpResponseMessage> DirectPut(string requestUri, byte[] bytes
             return await plot.PutAsync(requestUri, Sign(bytes), token);
         }
 
+        public byte[] GetPublicKey() => SignInfo.GetPublicKey();
+        public byte[] GetPrivateKey() => SignInfo.GetPrivateKey();
         public byte[] Sign(byte[] bytes) => [.. SignInfo.GetSignature(bytes), .. bytes];
+        //public byte[] Sign2(byte[] bytes) => [.. SignInfo.GetSignature(bytes)];
```
freddyrios, line 205: cdzc  ocd

CA_DataUploaderLib/ServerUploader.cs
```diff
@@ -792,6 +807,7 @@ public Signing(string loopName)
             }
 
             public byte[] GetPublicKey() => _rsaWriter.ExportCspBlob(false);
+            public byte[] GetPrivateKey() => _rsaWriter.ExportCspBlob(true);
```
freddyrios, line 810: do we need after the approach change at the end? (if removing please remove the callers)

freddyrios (approved):

CA_DataUploaderLib/ServerUploader.cs
LarsWithCA: removed

CA_DataUploaderLib/ServerUploader.cs
LarsWithCA: removed

PR #343, opened 2024-10-21
Expanded subsystems support

freddyrios:
- Introduced BaseSensorBoard.RunCustomCommand that allows the subsystem to send a custom command to the board
- Introduced RegisterBoardWriteActions overload that allows to have built in actions that are not based on a vector field
- Introduced IOconfOutput.GetExpandedInputConf protected overload that makes it easier for subclasses to implement GetExpandedInputConf
- re-introduced CustomWritesEnabled validation

### DIFF
--- a/CA_DataUploaderLib/BaseSensorBox.cs
+++ b/CA_DataUploaderLib/BaseSensorBox.cs
@@ -83,16 +83,14 @@ namespace CA_DataUploaderLib
             {
                 if (!map.IsLocalBoard)
                 {//if the board is not local we register the command validation with an empty action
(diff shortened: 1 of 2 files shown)


CA_DataUploaderLib/BaseSensorBox.cs
```diff
@@ -437,6 +448,12 @@ private async Task BoardReadLoop(MCUBoard board, string boxName, SensorSample.In
                     LogError(board, "Unexpected error on board read loop", ex);
                 }
             }
+
```
LarsWithCA, line 451: } is missing here

LarsWithCA (approved):

PR #346, opened 2024-11-06
improved a couple log messages + improved testability

freddyrios:
- Fixed: message about missing board with custom writes enabled was logged regardless of the setting
- removed the redundant "Stale sensor detected" that happens on some cases where report that we lost connection to a board / are attempting to reconnect
- time and logging dependencies are now injected to the CommandHandler and MCUBoard. BaseSensorBox now uses these dependencies via the CommandHandler.
- the serial port connection dependency is now injected to the MCUBoard
- added IOconfFile ctor overload that does notuse the static IOconfFileLoader.Loader

### DIFF
--- a/CA_DataUploaderLib/ICALogger.cs
+++ b/CA_DataUploaderLib/ICALogger.cs
@@ -1,0 +1,12 @@
+﻿using System;
+
+namespace CA_DataUploaderLib
+{
+    public interface ICALogger
+    {
+        void LogData(LogID id, string msg);
+        void LogError(LogID id, string msg, Exception ex);
+        void LogError(LogID id, string msg);
+        void LogInfo(LogID id, string msg);
+    }
+}
--- a/CA_DataUploaderLib/MCUBoard.cs
+++ b/CA_DataUploaderLib/MCUBoard.cs
@@ -594,26 +594,17 @@ namespace CA_DataUploaderLib
             }
         }
 
-        public class Dependencies(
-            TimeProvider timeProvider, Action<LogID, string> logInfo, Action<LogID, string> logError, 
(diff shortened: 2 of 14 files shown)


CA_DataUploaderLib/ICALogger.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+
+namespace CA_DataUploaderLib
+{
+    public interface ICALogger
```
LarsWithCA, line 5: why not just `ILogger` ?

CA_DataUploaderLib/MCUBoard.cs
```diff
@@ -600,15 +594,80 @@ internal void CopyTo(MCUBoard mcuBoard)
             }
         }
 
-        public class HeaderDependencies(TimeProvider timeProvider, Action<LogID, string> logInfo, Action<LogID, string> logError, Action<LogID, string, Exception> logException)
+        public class Dependencies(
+            TimeProvider timeProvider, Action<LogID, string> logInfo, Action<LogID, string> logError,
```
LarsWithCA, line 598: How about giving an ILogger instead of 4 different actions?

LarsWithCA:
Awesome stuff 👍

CA_DataUploaderLib/ICALogger.cs
freddyrios: used Ilog to not use the same name as Microsoft's one

CA_DataUploaderLib/MCUBoard.cs
freddyrios: done

LarsWithCA (approved):

LarsWithCA (approved):

Issue #349, opened 2024-11-11
Handle increased read delays when board is writing to flash


LarsWithCA:
When this is about to happen, this status bit is raised: https://github.com/copenhagenatomics/CA_Embedded_Libraries/blob/a051d1b2a37c753dce25f7406e9136cc9c1eece6/STM32/Util/Inc/systemInfo.h#L33

Relevant code: `BaseSensorBox.ReadSensors`

PR #350, opened 2024-11-18
Handle increased read delays when board is writing to flash

LarsWithCA:
https://github.com/copenhagenatomics/CA_DataUploader/issues/349

### DIFF
--- a/CA_DataUploaderLib/BaseSensorBox.cs
+++ b/CA_DataUploaderLib/BaseSensorBox.cs
@@ -487,6 +487,18 @@ namespace CA_DataUploaderLib
                         ProcessLine(numbers, board, targetSamples);
                         lastValidReadTime = _cmd.Time.GetTimestamp();
                         _boardsState.SetState(boxName, ConnectionState.ReceivingValues);
+
+                        if (status != _lastStatus && (status & 0x01000000) != 0) // Flash ongoing?
+                            timeBetweenReads = TimeSpan.FromSeconds(1.5);
+                        if (status != _lastStatus && (status & 0x01000000) == 0 && (_lastStatus & 0x01000000) != 0) // Flash done?
+                            timeBetweenReads = TimeSpan.FromMilliseconds(board.ConfigSettings.MillisecondsBetweenReads * 1.5);
+                        if (status != _lastStatus && (status & 0x80000000) != 0) // Error?
+                        {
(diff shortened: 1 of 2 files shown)


freddyrios (approved):

CA_DataUploaderLib/BaseSensorBox.cs
```diff
@@ -487,21 +487,25 @@ private async Task ReadSensors(MCUBoard board, string boxName, SensorSample.Inpu
                         ProcessLine(numbers, board, targetSamples);
                         lastValidReadTime = _cmd.Time.GetTimestamp();
                         _boardsState.SetState(boxName, ConnectionState.ReceivingValues);
+
+                        if (status != _lastStatus && (status & 0x01000000) != 0) // Flash ongoing?
```
matias-cphatomics, line 491: Do you need the comparison `!=0` here? I think it would evaluate to 1 if it is set. Same for the last comparison in line 493. 

Otherwise it looks good :=)

LarsWithCA:
@matias-cphatomics could you prepare a board with the modified firmware so I can test the change?

matias-cphatomics (approved):

CA_DataUploaderLib/BaseSensorBox.cs
LarsWithCA: it's either that or casting/converting it to a bool - I like this better :)

matias-cphatomics: Wouldn't an int evaluating to anything but 0 pass as true? That's how it is in C, but maybe it is different for you. Anyway also not something I'm really concerned with just a suggestion. :smile:

LarsWithCA: different in C# :)
image

LarsWithCA:
@freddyrios @matias-cphatomics I tested the oxygen board that Matias gave me. Without the changes in this PR, I get such lines in the B.log:

> 11.19 12:43:15.386 - No data available (False) - Generic - Luminox ZrO2Oxygen 1800124650501620313232 COM3
11.19 12:43:15.927 - Time from no data available to data available: 538,1263ms - Generic - Luminox ZrO2Oxygen 1800124650501620313232 COM3

With the changes in this PR, this doesn't happen.

PR #358, opened 2024-12-17
fixed: system can run out of memory with high events rate

freddyrios:
fixed: system can run out of memory when events are generated at a higher rate than they are processed

The most likely scenario is logging in cases of non throttled recurring failures that report errors at a high rate without any throttling), combined with distributed hosts that throttle the processing rate of events gathered from subsystems via CommandHandler.DequeueEvents.

Other changes:
- added the log type to the max 2 messages in 5 minutes message to make it clearer that it is a limit x type of message

### DIFF
--- a/CA_DataUploaderLib/CommandHandler.cs
+++ b/CA_DataUploaderLib/CommandHandler.cs
@@ -33,8 +33,8 @@ namespace CA_DataUploaderLib
         private readonly TaskCompletionSource _runningTaskTcs = new(TaskCreationOptions.RunContinuationsAsynchronously);
         private readonly bool _isMultipi;
         private readonly List<ChannelWriter<DataVector>> _receivedVectorsWriters = [];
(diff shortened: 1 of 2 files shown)


CA_DataUploaderLib/CommandHandler.cs
```diff
@@ -280,20 +280,24 @@ public void OnNewVectorReceived(DataVector args)
...
         }
 
         /// <summary>
         /// Returns a (new) list of dequeued events.
         /// </summary>
         /// <param name="max">The maximum number of events to dequeue.</param>
+        /// <remarks>
+        /// If events are created at a higher rate than the host processes some events will never been returned by this method.
+        /// Subsystems normally throttle the error generation rate, but if some code path not doing that
+        /// sees repeating errors that causes more than 1k to be pending, the oldest events are dropped/skipped.
```
LarsWithCA, line 293: 1k - isn't it only 200?

PR #370, opened 2025-04-01
The host is now responsible of initializing the alerts instance

LarsWithCA:


freddyrios (approved):
re title: its more like the host is now responsible of initializing the alerts instance as any other subsystem

In this regards the CA_DataUploader.Program should also be initializing it.

PR #371, opened 2025-04-03
Simplified configuration approach by use of tags

freddyrios:
Simplified configuration approach by use of tags

Breaking Changes:
- IOconfRow.ToList no longer returns a copy on each call (and no longer reparses the line on each call)
- IOconfRow ctor no longer sets the Name property when requireName is false. Derived classes must explicitely set a Name instead.
- IOconfFileLoader is no longer public. By not having this public, we can change some of the underlying details without further breaking changes. Use IOconfFile.DefaultLoader, IOconfFile.DefaultFileExists and IOconfFile.WriteToDisk instead.

New features:
- Tag any config line by adding at the end of the lists the 'tags:' prefix and a space separated list. For example, in the below myheater has 2 tags: ovenheaters and phase
   `Heater;myheater;myacbox;02;850;tags:ovenheaters phase2`
- Pass a fields list by specifying tagfields:mytag;suffix:mysuffix in config lines that expect a list of fields. For example, the below sends a list of all the onoff field of all the heaters that has the tag ovenheaters:   Code;myplugin;0.10.0
   `myplugin;heaters_onoff;tagfields:ovenheaters;suffix:onoff`
- The Expand config line can add a row for each tagged item:
   `Expand;tagfields:mytag;suffix:mysuffix;row with $name`
- The Expand config line can also do it with a hardcoded list:
   `Expand;name1;name2;...;namen;-;row with $name`

Other changes:
- Added IOconfRow.ExpandsTags: derived classes can set it to true to enable tagsfields
- Changed IOconfRow.ValidateNameRegex to use the regex code generator.


CA_DataUploaderLib/IOconf/IOconfRow.cs
```diff
@@ -6,23 +6,38 @@
...
+
             if (!isUnknown && list[0] != type) throw new Exception("IOconfRow: wrong format: " + row);
             Type = list[0];
-            Name = list.Count >= 2  // could be overwritten elsewhere. 
+            if (requireName)
+                Name = list.Count >= 2 // could be overwritten elsewhere. 
                 ? list[1]
-                : requireName ? throw new Exception("IOconfRow: missing Name") : string.Empty;
+                : throw new Exception("IOconfRow: missing Name");
+            ExpandsTags = isUnknown;
```
LarsWithCA, line 34: is this due to Code-lines?

freddyrios: Yes

CA_DataUploaderLib/IOconf/IOConfFileLoader.cs
```diff
@@ -33,8 +33,12 @@ public static IEnumerable<IOconfRow> ParseLines(IIOconfLoader loader, IEnumerabl
...
             var linesList = lines.Select(x => x.Trim()).ToList();
             // remove empty lines and commented out lines
-            var lines2 = linesList.Where(x => !x.StartsWith("//") && x.Length > 2).ToList();
-            return lines2.Select(x => CreateType(loader, x, linesList.IndexOf(x) + 1));
+            var lines2 = linesList.Where(x => !x.StartsWith("//") && x.Length > 2).Select((x,i) => (row: x,line: i)).ToList();
+            var rows = lines2.Select(x => CreateType(loader, x.row, x.line + 1)).ToList();
+            var tags = rows.SelectMany(r => r.Tags.Select(t => (tag: t, rowname: r.Name))).ToLookup(r => r.tag, r=> r.rowname);
+            foreach (var row in rows)
+                row.UseTags(tags);
+            return rows.Concat(rows.SelectMany(r => r.GetExpandedConf().Select(l => CreateType(loader, l, r.LineNumber))));
```
LarsWithCA, line 41: I guess we will need two "versions" of an IO.conf that gets loaded by LoopControl:

- an original/unexpanded version, e.g. shown in WebUI + used when renaming fields.
- and an expanded version, which is the one LoopControl will actually use.

CA_DataUploaderLib/IOconf/IOconfRow.cs
LarsWithCA: please add a comment

CA_DataUploaderLib/IOconf/IOconfExpand.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Linq;
+
+namespace CA_DataUploaderLib.IOconf
+{
+    public sealed class IOconfExpand : IOconfRow
+    {
+        public const string ConfigName = "Expand";
+        public IOconfExpand(string row, int lineNum) : base(row, lineNum, ConfigName, false, false)
+        {
+            Format = $"{ConfigName};name1;name2;...;namen;-;rowstartshere;$name{Environment.NewLine}{ConfigName};tagfields:mytag;suffix:mysuffix;rowstartshere;$name";
```
LarsWithCA, line 12: only "-" in the "first version"?

CA_DataUploaderLib/IOconf/IOconfExpand.cs
```diff
@@ -0,0 +1,30 @@
...
+            ExpandsTags = true;
+            var list = ToList();
+            if (list.Count < 3 || string.IsNullOrEmpty(list[2]))
+                throw new FormatException($"Wrong format: {Row}. {Format}");
```
LarsWithCA, line 16: newline before {Format}?

CA_DataUploaderLib/IOconf/IOConfFileLoader.cs
LarsWithCA: maybe it already works like that?

CA_DataUploaderLib/IOconf/IOconfExpand.cs
LarsWithCA: row with $name

LarsWithCA (approved):

CA_DataUploaderLib/IOconf/IOconfExpand.cs
freddyrios: > only "-" in the "first version"?

Yes. Discussed offline: '-' is needed to be able to tell when a list ends within a line (as they use the same ; separator as any other config value). For tag fields its not needed as its already possible to tell where the row to expand begins. The tagfields feature adds a - for any list expanded that is not at the end of the line.

CA_DataUploaderLib/IOconf/IOConfFileLoader.cs
freddyrios: added IIOconf.GetEntriesWithoutExpansion that can has the original list of rows

LarsWithCA (approved):

LarsWithCA (approved):

PR #377, opened 2025-05-02
Made GenericOutput able to reference multiple target fields in the command template

LarsWithCA:
Make GenericOutput + LastAction able to handle multiple targets / target values.

Breaking changes:
- The signature of `BaseSensorBox.RegisterBoardWriteActions` has been modified to work with multiple target names and target values.
- The LastAction has been modified to work with indices (into the data vector) instead of individual values (from the data vector).

### DIFF
--- a/CA_DataUploaderLib/LastAction.cs
+++ b/CA_DataUploaderLib/LastAction.cs
@@ -10,42 +10,42 @@ namespace CA_DataUploaderLib
         private readonly TimeProvider timeProvider;
         private long lastActionExecutedTime;
 
-        public LastAction(double target, int repeatMilliseconds) : this(target, repeatMilliseconds, TimeProvider.System) { }
-        public LastAction(double target, int repeatMilliseconds, TimeProvider timeProvider) : this([target], repeatMilliseconds, timeProvider) { }
-        public LastAction(IEnumerable<double> targets, int repeatMilliseconds) : this(targets, repeatMilliseconds, TimeProvider.System) { }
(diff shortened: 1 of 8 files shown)


freddyrios (approved):

CA_DataUploaderLib/LastAction.cs
```diff
@@ -34,7 +33,7 @@ public LastAction(IEnumerable<int> targetIndices, double defaultValue, int repea
         /// Note however that there are many other ways that bit flips could affect related actuations and also the functioning of this very class.
         /// </remarks>
         public bool ChangedOrExpired(double[] newVector, DateTime currentVectorTime) =>
-            Indices.Any(i => Vector[i] != newVector[i]) || (repeatMilliseconds > -1 && timeProvider.GetElapsedTime(lastActionExecutedTime).TotalMilliseconds >= repeatMilliseconds) || currentVectorTime >= TimeToRepeat;
+            Indices.Any(i => Vector.Length == 0 || Vector[i] != newVector[i]) || (repeatMilliseconds > -1 && timeProvider.GetElapsedTime(lastActionExecutedTime).TotalMilliseconds >= repeatMilliseconds) || currentVectorTime >= TimeToRepeat;
```
freddyrios, line 36: a bit confused how Vector length 0 can happen

if the check is needed it could move out of the any

freddyrios (approved):

CA_DataUploaderLib/LastAction.cs
LarsWithCA: the LastAction is created with an empty Vector

PR #378, opened 2025-05-06
Added support for OvenArea

freddyrios:
The changes fixes the loop name of the single node not matching changes in a separate ioconf loaded in memory.

`Format: OvenArea;AreaNumber e.g. OvenArea;2`

This allows to introduce an oven area without requiring x heater Oven lines. This allows to use oven areas in plugins that have their own oven control logic.

Heaters lines that dont use the built in Oven logic must not specify a MaxTemperature e.g. the heater line should look like this: Heater;myheater;myacboard;1

This also means there is no longer a warning for missing Oven lines for a heater.

Breaking changes:
- `IOconfNode.GetSingleNode` is no longer public. Use `CommandHandler.GetNodes` instead. It gets the list of all nodes and in the case of a single deployment the single node.

Other changes:
- heater decision tests now run at a higher level, using more of the full system logic to run
- improved some of the duplicate configurations messages

### DIFF
--- a/CA_DataUploaderLib/CommandHandler.cs
+++ b/CA_DataUploaderLib/CommandHandler.cs
@@ -257,10 +257,10 @@ namespace CA_DataUploaderLib
         }
 
         private IOconfNode GetCurrentNode() => GetNodes().Single(n => n.IsCurrentSystem);
-        private List<IOconfNode> GetNodes()
+        public List<IOconfNode> GetNodes()
         {
             var nodes = _ioconf.GetEntries<IOconfNode>().ToList();
-            return nodes.Count > 0 ? nodes : [IOconfNode.GetSingleNode(_ioconf.GetLoopName())];
+            return nodes.Count > 0 ? nodes : [IOconfNode.GetSingleNode(_ioconf)];
         }
 
         public void OnNewVectorReceived(DataVector args)
--- a/CA_DataUploaderLib/IOconf/IOConfFileLoader.cs
+++ b/CA_DataUploaderLib/IOconf/IOConfFileLoader.cs
@@ -27,9 +27,6 @@ namespace CA_DataUploaderLib.IOconf
 
         public static (List<IOconfRow> original, List<IOconfRow> expanded) ParseLines(IIOconfLoader loader, IEnumerable<string> lines)
         {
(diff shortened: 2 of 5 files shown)


CA_DataUploaderLib/IOconf/IOconfOvenArea.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+    internal sealed class IOconfOvenArea : IOconfDriver, IIOconfRowWithDecision
+    {
+        public IOconfOvenArea(string row, int lineNum) : base(row, lineNum, "OvenArea")
+        {
+            Format = "OvenArea;AreaNumber";
+
+            var list = ToList();
+            if (!int.TryParse(list[1], out OvenArea)) 
+                throw new Exception($"Oven are must be a number: {row} {Format}");
```
LarsWithCA, line 17: are => area

CA_DataUploaderLib/IOconf/IOconfOvenArea.cs
```diff
@@ -0,0 +1,32 @@
...
+            if (OvenArea < 1)
+                throw new Exception("Oven area must be a number bigger or equal to 1");
```
LarsWithCA, line 19: bigger than or

UnitTests/FullDecisionTestContext.cs
```diff
@@ -0,0 +1,80 @@
...
+        private readonly ChannelReader<string> logs;
+        public DateTime InitialTime { get; }
+
+        public FullDecisionTestContext(string config, Dictionary<string, double>? initialVectorSamples = null)
+        {
+            var loader = new IOconfLoader();
+            Subsystems.RegisterIOConfAndThirdPartyBoardsProtocols(loader);
+            var ioconf = new IOconfFile(loader, config.SplitNewLine());
+            var logger = new ChannelLogger();
+            cmd = new CommandHandler(ioconf, runCommandLoop: false, logger: logger);//TODO: time provider and logging?
```
LarsWithCA, line 27: 3x TODOs in this file

LarsWithCA (approved):

LarsWithCA (approved):

PR #405, opened 2025-11-10
Fix for Issue #321: TimeAfter overflow for values > 596 hours

### LINKED ISSUE #321: plugin time transitions above 596 hours don't work correctly
Reason: underneath there are some methods taking milliseconds for time transitions, which use int that has a max value of 2147483647

Specifically at CA.LoopControlPluginBase.DataVector.TimeAfter

Modify or add overload (for backward compatibility)? public double TimeAfter(int milliseconds)

YannickBrot:
Resolves #321
Changed TimeAfter parameter from int to long to prevent overflow, which also 
maintains backward compatibility since int values can be implicitly converted 
to long.

Updated all 3 method calls to cast to long instead of int, as this is where 
the overflow occurs before the method is called.

Changed project reference in CA_DataUploaderLib.csproj from package to project 
reference to allow the build to run with the updated CA.LoopControlPluginBase.

Added unit tests.

### DIFF
--- a/CA.LoopControlPluginBase.Tests/DataVectorTests.cs
+++ b/CA.LoopControlPluginBase.Tests/DataVectorTests.cs
@@ -34,7 +34,9 @@ namespace CA.LoopControlPluginBase.Tests
             var vector = new DataVector(baseTime, data);
 
             // Act - tests the old int overload still works (backward compatibility)
+            #pragma warning disable CS0618
             var result = vector.TimeAfter(int.MaxValue);
+            #pragma warning restore CS0618
 
             // Assert
             var expectedTime = baseTime.AddMilliseconds(int.MaxValue);


freddyrios (changes requested):
Hi @YannickBrot, thanks for contributing to the project!

CA.LoopControlPluginBase is a PluginContract library. Plugins depend on this thin library, so that unrelated (breaking) changes to the host do not affect them. There are some examples of plugins in the zip file of this repo: https://github.com/copenhagenatomics/DefaultPlugins/releases/tag/v1.2.0.

This is why we manage its lifecycle separately and use a package reference for it. For example, this is the last time it changed: https://github.com/copenhagenatomics/CA_DataUploader/releases/tag/plugins1.2.0.

Would be great if you can:
- Instead of changing the method argument in place for TimeAfter, add a separate overload. It avoids breaking existing plugins due to the [binary breaking change due of changing the argument type change](https://learn.microsoft.com/en-us/dotnet/core/compatibility/library-change-rules#members). You can add an Obsolete attribute to the old method.
- Move the CA.LoopControlPluginBase changes to a separate PR. In this way, we can make a release of the library with the new method, so that you can use it in the current PR.

YannickBrot:
Thanks for the feedback! I didn't consider the binary compatibility issue 
with plugins, that makes sense. 

I've updated the PR to add an overload instead of changing the signature. 
The old TimeAfter(int) is still there but marked as Obsolete, and I added 
the new TimeAfter(long) overload.

I've also added unit tests for the new overload, but they're failing to 
build because UnitTests references CA_DataUploaderLib, which has a package 
reference to CA.LoopControlPluginBase version 1.2.0. This means the tests 
can't see the new TimeAfter(long) overload from the local project.

How do you want to handle this?

freddyrios:
@YannickBrot correct, the main problem is that we only currently have tests for the main assembly and changing those triggers the "main build" instead of only the "build plugins" build. We'll push a change to get a test project for it in place.

PR #409, opened 2025-11-20
Allow loading IO.conf and CodeRepoURLs.json from a specific folder.

LarsWithCA:
- Allow loading IO.conf and CodeRepoURLs.json from a specific folder.
- CommandHandler: get a user-friendly error when having two decisions with the same name.
- DULutil.ExecuteCommand: added newlines to error output.

### DIFF
--- a/CA_DataUploaderLib/Helpers/DULutil.cs
+++ b/CA_DataUploaderLib/Helpers/DULutil.cs
@@ -94,7 +94,7 @@ namespace CA_DataUploaderLib.Helpers
                     WorkingDirectory = Environment.CurrentDirectory
                 }
             };
-            p.ErrorDataReceived += (sender, e) => { errorOutput += (string.IsNullOrEmpty(errorOutput) ? string.Empty : Environment.NewLine) + e.Data; };
+            p.ErrorDataReceived += (sender, e) => { errorOutput += e.Data + Environment.NewLine; };
             if (!p.Start())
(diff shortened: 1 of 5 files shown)


CA_DataUploaderLib/Helpers/DULutil.cs
```diff
@@ -94,17 +94,17 @@ public static (string stdOutput, string errOutput, int exitCode) ExecuteCommand(
                     WorkingDirectory = Environment.CurrentDirectory
                 }
             };
-            p.ErrorDataReceived += (sender, e) => { errorOutput += e.Data; };
```
freddyrios: why not += e.Data + Environment.NewLine? the event docs say "The [ErrorDataReceived](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.process.errordatareceived?view=net-7.0#system-diagnostics-process-errordatareceived) event indicates that the associated [Process](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.process?view=net-7.0) has written a line that's terminated with a newline (carriage return (CR), line feed (LF), or CR+LF) to its redirected [StandardError](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.process.standarderror?view=net-7.0) stream.". 

FWIW its a bit scary that a non line terminated error would not end up being visible if they are strict about that behavior

PR #421, opened 2026-05-12
Reduce data loss during upload outages

YannickBrot:
This PR closes #1338.

It increases the vector upload buffer capacity from `10000` to `20000`. With the default main loop delay of `100 ms`, that is roughly `10000 * 100 ms = 1000 s = 16.7 min` before, and `20000 * 100 ms = 2000 s = 33.3 min` after.

The memory increase depends on the number of uploaded fields. As a rough example, with about 1000 uploaded `double` values per vector, the raw vector payload capacity goes from about `80 MB` to about `160 MB`, before normal object/array overhead.

Previously, queued vectors were dequeued before upload. If the HTTP upload failed, those vectors had already been removed from the queue and were discarded.

This PR keeps the bounded `Channel<DataVector>` approach and adds a small pending retry batch:
- vector channel capacity is `20000`
- vector upload batch size is `600` vectors, roughly 1 minute at the default 10 Hz loop
- failed vector batches remain pending and are retried before newer channel data is dequeued
- successful vector uploads clear the pending batch
- event uploads now use the same retry policy with one pending event retried before newer events

Retry policy (shared by vectors and events):
- **Retried indefinitely** — HTTP/network failures with no status code (e.g. connection failures, `HttpRequestException` without a `StatusCode`) and HTTP status codes `404`, `408`, `429`, `502`, `503`, `504`.
- **Non-retryable, dropped after 3 attempts** — all other HTTP status codes (e.g. `500`) and non-HTTP local exceptions (e.g. JSON/serialization errors). After 3 attempts the failing batch/event is logged and discarded so one bad batch cannot block uploads forever.

Tests added for:
- vector batch dequeue limit
- pending vector batch retry ordering
- empty vector queue behavior
- shared retryable/non-retryable HTTP status policy, including `500` as non-retryable
- pending event retry/drop behavior for retryable and non-retryable failures


CA_DataUploaderLib/ServerUploader.cs
```diff
@@ -476,7 +479,11 @@ async Task<bool> Post()
                 {
                     using var response = await Plot.PostVectorAsync(buffer, timestamp);
                     var success = CheckAndLogFailures(response, "Failed posting vector");
-                    stateWriter.TryWrite(UploadState.UploadedVector);
+                    if (success)
+                    {
+                        _vectorsBuffer.RemoveThrough(snapshot.LastSequence);
+                        stateWriter.TryWrite(UploadState.UploadedVector);
+                    }
```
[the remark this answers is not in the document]
YannickBrot: We dont want to block writers while uploading as we want to favor the newest data. The snapshot is a local variable inside PostQueuedVectorAsync(). Once the HTTP request returns or throws, that method finishes and the snapshot is no longer referenced by our code. HTTPClient.Timeout is 100 sec so worst case will be: 20.000 (full buffer) + 100 (seconds request can be in flight) * 10 (hz) = 21.000 - so not really an issue I think.

CA_DataUploaderLib/ServerUploader.cs
```diff
@@ -476,13 +505,20 @@ async Task<bool> Post()
...
+                        return true;
+                    }
+
+                    return HandleFailedVectorUpload(response.StatusCode, pendingVectors);
                 }
                 catch (HttpRequestException ex)
                 {
                     OnError($"Failed posting vector (HttpStatusCode={ex.StatusCode}, HttpRequestError={ex.HttpRequestError}).", ex);
-                    return false;
+                    return ShouldRetryUpload(ex.StatusCode) ? false : HandleFailedVectorUpload(ex.StatusCode, pendingVectors);
```
freddyrios: ShouldRetryUpload is already checked by HandleFailedVectorUpload

```suggestion
                    return HandleFailedVectorUpload(ex.StatusCode, pendingVectors);
```

CA_DataUploaderLib/ServerUploader.cs
```diff
@@ -503,84 +558,132 @@ private IEnumerable<EventFiredArgs> FilterEvents(IEnumerable<EventFiredArgs> eve
...
+            RemoveExpiredDuplicateEvents();
+
+            if (pendingEvent != null)
+            {
+                pendingEvent = await PostOrKeepPendingEvent(pendingEvent);
+                if (pendingEvent != null)
+                    return pendingEvent;
+            }
+
+            var events = DequeueAllEntries(_eventsChannel.Reader);
```
freddyrios: DequeueAllEntries is problematic with the current logic, as events get dequeued but we only keep the first pendingEvent that fails (the rest are discarded)

Because we don't batch them, one option is to handle the retries within the foreach below.
We might want to do it after the TrackDuplicate call, because currently I think that would track the post retries as if we had received duplicate events

freddyrios (approved):

PR #422, opened 2026-05-20
Fix IO.conf data loss during power cut (#1370)

YannickBrot:
Fixes #1370.

Reproduced the bug reliably by adding 3.5k comments to IO.conf and pulling the plug about 5 sec after, both IO.conf and IO.conf.backup ended up empty. So the existing backup mechanism was not actually protecting us: we had a second issue where IO.conf.backup would also be empty, not just IO.conf.

The fix:
- **Flushes (fsync) after making the backup**, so once we move on we are guaranteed never to lose the previous IO.conf even if power dies mid-write of the new one.
- **Writes the new IO.conf via a temp file in the same directory** (instead of Path.GetTempFileName() which lives in /tmp). /tmp can be a different filesystem depending on configuration, which turns File.Move into a copy+delete rather than an atomic rename.
- **fsyncs the temp file before the rename**, so the new contents are physically on the disk/card before the directory entry changes.

### DIFF
--- a/CA_DataUploaderLib/IOconf/IOConfFileLoader.cs
+++ b/CA_DataUploaderLib/IOconf/IOConfFileLoader.cs
@@ -61,7 +61,7 @@ namespace CA_DataUploaderLib.IOconf
                 while (File.Exists(newFilename));
 
                 File.Copy(filename, newFilename);
-                using var backupFs = new FileStream(newFilename, FileMode.Open, FileAccess.ReadWrite);
+                using var backupFs = new FileStream(newFilename, FileMode.Open, FileAccess.Write);
                 backupFs.Flush(flushToDisk: true);
             }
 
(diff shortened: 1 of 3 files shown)


CA_DataUploaderLib/IOconf/IOConfFileLoader.cs
```diff
@@ -61,10 +61,19 @@ public static void WriteToDisk(string ioconf)
                 while (File.Exists(newFilename));
 
                 File.Copy(filename, newFilename);
+                using var backupFs = new FileStream(newFilename, FileMode.Open, FileAccess.ReadWrite);
+                backupFs.Flush(flushToDisk: true);
+            }
+
+            var temp = filename + ".tmp";
+            using (var fs = new FileStream(temp, FileMode.Create, FileAccess.Write))
```
LarsWithCA, line 69: FileMode.CreateOrOverwrite

LarsWithCA:
Please flip the order of write operations in IOconfFile.WriteToDisk() + also flush IOconfCodeRepo.WriteURLsToFile

CA_DataUploaderLib/IOconf/IOConfFileLoader.cs
YannickBrot: FileMode.Create already overwrites if already exists :)

LarsWithCA (approved):

